PR #30, opened 2016-08-19
Add custom user location annotation example

friedbunny:
Adds a custom user location example.
- Does not attempt to handle accuracy or pitch.
- I would like to have a permanent heading indicator, but this depends on https://github.com/mapbox/mapbox-gl-native/issues/5523.
### Todo
- [x] Objective-C
- [ ] Cleanup and review.

/cc @1ec5 @frederoni


friedbunny:
→ #50

PR #65, opened 2017-02-14
Blocking map gestures delegate example

fabian-guerra:


Examples/Swift/BlockingGesturesDelegateExample.swift
```diff
@@ -0,0 +1,57 @@
...
+        // Denver, Colorado
+        let center = CLLocationCoordinate2D(latitude: 39.748947, longitude: -104.995882)
+        
+        // Starting point.
+        mapView.setCenter(center, zoomLevel: 10, direction: 0, animated: false)
+        
+        // Colorado's bounds
+        let ne = CLLocationCoordinate2D(latitude: 40.989329, longitude: -102.062592)
+        let sw = CLLocationCoordinate2D(latitude: 36.986207, longitude: -109.049896)
+        colorado = MGLCoordinateBoundsMake(sw, ne)
```
1ec5: Use `MGLCoordinateBounds(sw:ne:)` instead of `MGLCoordinateBoundsMake(_:_:)`.

boundsj:
I think it'd be interesting to do a performance comparison with the approach already in this PR and the approach outlined in https://github.com/mapbox/ios-sdk-examples/pull/65#pullrequestreview-21907456. A less accurate, limited solution may still be useful for cases involving offline maps that usually involve higher zoom levels.

Examples/ObjectiveC/BlockingGesturesDelegateExample.m
```diff
@@ -0,0 +1,56 @@
...
+
+@implementation BlockingGesturesDelegateExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds];
+    mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    mapView.rotateEnabled = NO;
+    mapView.minimumZoomLevel = 10;
```
1ec5: Is it still necessary to disable rotation and cap the zoom level, now that you’re doing the set-and-revert dance?

Examples/ObjectiveC/BlockingGesturesDelegateExample.m
```diff
@@ -0,0 +1,56 @@
...
+    [self.view addSubview:mapView];
+}
+
+- (BOOL)mapView:(MGLMapView *)mapView shouldChangeFromCamera:(MGLMapCamera *)oldCamera toCamera:(MGLMapCamera *)newCamera
+{
+    // Get current coordinates
+    CLLocationCoordinate2D newCameraCenter = newCamera.centerCoordinate;
+    MGLMapCamera *camera = mapView.camera;
+    
+    // Get new bounds
```
1ec5: These comments say _what_ the developer is expected to do, but not _why_. An example needs to explain the rationale behind each non-obvious step, in order to avoid the phenomenon of [cargo cult programming](http://www.catb.org/~esr/jargon/html/C/cargo-cult-programming.html). In particular, we should be explicit about the fact that we’re changing and resetting the map view’s camera in order to predict the extents of the new viewport.

Examples/ObjectiveC/BlockingGesturesDelegateExample.m
```diff
@@ -0,0 +1,56 @@
...
+    mapView.camera = newCamera;
+    MGLCoordinateBounds newVisibleCoordinates = mapView.visibleCoordinateBounds;
+    mapView.camera = camera;
+    
+    // Test if the new camera center point and boundaries are inside colorado
+    BOOL inside = MGLCoordinateInCoordinateBounds(newCameraCenter, self.colorado);
+    BOOL intersects = MGLCoordinateInCoordinateBounds(newVisibleCoordinates.ne, self.colorado) && MGLCoordinateInCoordinateBounds(newVisibleCoordinates.sw, self.colorado);
```
1ec5: Ah, I see why you’ve restricted the zoom level and disabled rotation. Can you think of a way to test whether the new viewport would fit inside the allowed region (Colorado) while accounting for rotation? Perhaps the logic in [turf-inside](https://github.com/Turfjs/turf/tree/master/packages/turf-inside) would be helpful here, although it might be overkill.

fabian-guerra: I'm already taking a similar approach like function `inBBox` and also accounts for restricting rotation if its out of boundaries.

Examples/ObjectiveC/BlockingGesturesDelegateExample.m
fabian-guerra: I changed the comments to make it more explicit.

Examples/ObjectiveC/BlockingGesturesDelegateExample.m
fabian-guerra: The example I made deals with rotation but I considered was not important for restricting boundaries, I reverted this.

fabian-guerra:
@1ec5 @captainbarbosa @jmkiley could you please take a look.

Examples/Swift/BlockingGesturesDelegateExample.swift
```diff
@@ -0,0 +1,60 @@
...
+        mapView.camera = newCamera
+        let newVisibleCoordinates = mapView.visibleCoordinateBounds
+        
+        // Revert the camera
+        mapView.camera = currentCamera
+
+        // Test if the newCameraCenter and newVisibleCoordinates
+        // are inside self.colorado
+        let inside: Bool = MGLCoordinateInCoordinateBounds(newCameraCenter, self.colorado)
+        let intersects: Bool = MGLCoordinateInCoordinateBounds(newVisibleCoordinates.ne, self.colorado) && MGLCoordinateInCoordinateBounds(newVisibleCoordinates.sw, self.colorado)
```
1ec5: Nit: drop `self.`, since it’s implied.

Examples/ObjectiveC/BlockingGesturesDelegateExample.m
```diff
@@ -0,0 +1,64 @@
...
+    // Colorado's bounds
+    CLLocationCoordinate2D ne = CLLocationCoordinate2DMake(40.989329, -102.062592);
+    CLLocationCoordinate2D sw = CLLocationCoordinate2DMake(36.986207, -109.049896);
+    self.colorado = MGLCoordinateBoundsMake(sw, ne);
+    
+    [self.view addSubview:mapView];
+}
+
+// This example uses Colorado's boundaries to restrict
+// the camera movement
```
1ec5: The use of a period at the end of a comment is inconsistent in this file. Here there’s no period at the end of the sentence; above, “Starting point” does end with a period, even though it isn’t a sentence. Normally I end all the comments with periods if any of them is a sentence.

Examples/Swift/BlockingGesturesDelegateExample.swift
```diff
@@ -0,0 +1,60 @@
...
+    // the camera movement
+    
+    func mapView(_ mapView: MGLMapView, shouldChangeFrom oldCamera: MGLMapCamera, to newCamera: MGLMapCamera) -> Bool {
+        
+        // Get the current camera to restore it after
+        let currentCamera = mapView.camera
+        
+        // From the new camera obtain the center to test
+        // if it's inside the boundaries
+        let newCameraCenter: CLLocationCoordinate2D = newCamera.centerCoordinate
```
1ec5: Nit: `centerCoordinate`’s type is inferred as `CLLocationCoordinate2D`, so there’s no need to declare `newCameraCenter`’s type here. (Same with `inside` and `intersects` below.)

Examples/ObjectiveC/BlockingGesturesDelegateExample.m
1ec5: Yes, this code is fine. I was mistaken previously in thinking that this logic would produce incorrect results when the map is rotated, but the visible coordinate bounds will extend beyond the allowed region. This logic would not work if the map is tilted or if the allowed region itself is rotated.

PR #69, opened 2017-03-08
MGLShapeCollectionFeature example

captainbarbosa:
This example illustrates how [`MGLShapeCollectionFeature`](https://www.mapbox.com/ios-sdk/api/3.5.0-beta.2/Style%20Primitives.html#/c:objc(cs)MGLShapeCollectionFeature) can be used to add a dataset using the [Mapbox Datasets API](https://www.mapbox.com/api-documentation/#retrieve-a-dataset). Both circle and line layers to a map using runtime styling properties and predicates. This also takes advantage of new `circleStrokeWidth` and `circleStrokeColor` paint attributes introduced in [v3.5.0](https://www.mapbox.com/ios-sdk/api/3.5.0-beta.2/) of the Mapbox iOS SDK.

screen shot 2017-03-07 at 5 51 34 pm


Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,56 @@
...
+@interface ShapeCollectionFeatureExample () <MGLMapViewDelegate>
+
+@end
+
+@implementation ShapeCollectionFeatureExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame: self.view.bounds styleURL: [MGLStyle lightStyleURLWithVersion:9]];
```
1ec5: Nit: remove the spaces after the colons.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,56 @@
...
+    
+    [mapView setCenterCoordinate:CLLocationCoordinate2DMake(38.897435, -77.039679) zoomLevel:12 animated:NO];
+    mapView.delegate = self;
+    
+    [self.view addSubview:mapView];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    
+    NSURL *url = [[NSURL alloc] initWithString: @"https://api.mapbox.com/datasets/v1/mapbox/cj004g2ay04vj2xls3oqdu2ou/features?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpemc0YWlpNzAwcXUyd21ldDV6OWpxMGwifQ.A92RQZpwUgtGtCmdSE4-ow"];
```
1ec5: Use `+[NSURL URLWithString:]` for simplicity.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,56 @@
...
+    if (url) {
```
1ec5: It’s true that `+[NSURL URLWithString:]` and `-[NSURL initWithString:]` can return nil, but only if the string is a malformed URL. You know (and I’ve made sure) that the URL above is well-formed, so you don’t need to check for nil here.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
1ec5: Is it prudent to use default access token for the `mapbox` account in this example? Could we put in some `<#placeholder token#>`s for the access token and dataset ID instead? Or even replace this usage of the datasets API with a more straightforward use of a local file:

```objc
NSURL *url = [[NSBundle mainBundle] URLForResource:@"stations" withExtension:@"geojson"];
```

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,56 @@
...
+        NSData *data = [[NSData alloc] initWithContentsOfURL:url];
+        MGLShape *feature = [MGLShape shapeWithData:data encoding: NSUTF8StringEncoding error: nil];
```
1ec5: Nit: `error` is of type `NSError **`, which is not an Objective-C object pointer but rather a double-pointer. Use `NULL` instead of `nil` to indicate the nonexistence of a non-object pointer. See [this article](http://nshipster.com/nil/) for more nuance.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,56 @@
...
+        
+        // Create source and add it to the map style.
+        MGLShapeSource *source = [[MGLShapeSource alloc] initWithIdentifier:@"transit" shape:feature options:nil];
```
1ec5: Since you don’t need to manipulate the GeoJSON data before creating a shape source with it, you could instead call `-[MGLShapeSource initWithIdentifier:URL:options:]`. There are strong use cases for either approach, so use the approach you think is more important to teach in this example.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,56 @@
...
+        [style addSource:source];
+        
+        // Create station style layer.
+        MGLCircleStyleLayer *circleLayer = [[MGLCircleStyleLayer alloc] initWithIdentifier:@"stations" source:source];
+        circleLayer.predicate = [NSPredicate predicateWithFormat:@"%@" argumentArray:@[@"Station"]];
```
1ec5: This isn’t a valid predicate. A valid predicate must include an operator. Based on the data in the dataset, I suspect you want to check whether the `TYPE` attribute is `Station`. Here are some ways of expressing that:

```objc
[NSPredicate predicateWithFormat:@"TYPE = 'Station'"]
[NSPredicate predicateWithFormat:@"TYPE == 'Station'"]
[NSPredicate predicateWithFormat:@"TYPE = %@", @"Station"]
[NSPredicate predicateWithFormat:@"TYPE IN {'Station'}"]
[NSPredicate predicateWithFormat:@"TYPE IN %@", @[@"Station"]]
[NSPredicate predicateWithFormat:@"{'Station'} CONTAINS TYPE"]
```

Note that I’m using `+[NSPredicate predicateWithFormat:]` here instead of `+[NSPredicate predicateWithFormat:argumentArray:]`. The latter is useful if you’re given an array to put in the second parameter and would find it inconvenient to have to lump it in with the format in the first parameter.

Examples/Swift/ShapeCollectionFeatureExample.swift
```diff
@@ -0,0 +1,52 @@
...
+        
+        mapView.setCenter(CLLocationCoordinate2D(latitude:38.897435, longitude: -77.039679), zoomLevel: 12, animated: false)
+        
+        mapView.delegate = self
+        view.addSubview(mapView)
+    }
+    
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        
+        if let url = URL(string: "https://api.mapbox.com/datasets/v1/mapbox/cj004g2ay04vj2xls3oqdu2ou/features?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpemc0YWlpNzAwcXUyd21ldDV6OWpxMGwifQ.A92RQZpwUgtGtCmdSE4-ow"){
```
1ec5: As in Objective-C, you know the URL is well-formed, so you can `!` force-unwrap it.

Examples/Swift/ShapeCollectionFeatureExample.swift
```diff
@@ -0,0 +1,52 @@
...
+            let feature = try! MGLShape(data: data, encoding: String.Encoding.utf8.rawValue) as! MGLShapeCollectionFeature
+            
+            // Create source and add it to the map style.
+            let source = MGLShapeSource(identifier: "transit", shape: feature, options: nil)
+            style.addSource(source)
+            
+            // Create station style layer.
+            let circleLayer = MGLCircleStyleLayer(identifier: "stations", source: source)
+            // Use a predicate to filter out non-points.
+            circleLayer.predicate = NSPredicate(format: "TYPE = %@", ["Station"])
```
1ec5: This predicate says `TYPE` must be the array containing `Station`. We don’t support setting feature properties to arrays, and the dataset always sets `TYPE` to a simple string anyways. So you probably mean either `TYPE = 'Station'` or `TYPE IN {'Station'}`. I’m guessing you arrived at this syntax because you meant to use `NSPredicate(format:argumentArray:)`, but `NSPredicate(format:)` is preferable, as explained above.

install.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-MBGLVERSION='3.4.1'
+MBGLVERSION='3.4.2'
```
1ec5: Since we’re using `circleStrokeWidth` and `circleStrokeColor`, this should point to a v3.5.0 beta, right?

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
friedbunny: I’m onboard with `<#placeholder token#>`, if we’re intent on demoing the datasets API.

install.sh
friedbunny: Yeah, d42768cff5e817d94f6d87ecd0d1372972d127e7 in the `release-v3.5.0` should have covered this.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
jmkiley: We switched it to this method while troubleshooting something else. We can briefly explain in a comment what is being accomplished with that method, as well as the alternative.

install.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-MBGLVERSION='3.4.1'
+MBGLVERSION='3.5.0'
```
friedbunny: This won’t work until v3.5.0 is released. Let’s remove anything related to `install.sh` from this PR, as it shouldn’t be necessary.

friedbunny: > shouldn’t be necessary.

I should clarify “necessary”: the `release-v3.5.0` branch has the required changes to develop examples for that release. This PR should be (re)based on that branch, which eliminates the need to fiddle with `install.sh`.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,55 @@
...
+    mapView.delegate = self;
+    
+    [self.view addSubview:mapView];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    
+    NSURL *url = [NSURL URLWithString: @"https://api.mapbox.com/datasets/v1/mapbox/cj004g2ay04vj2xls3oqdu2ou/features?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpemc0YWlpNzAwcXUyd21ldDV6OWpxMGwifQ.A92RQZpwUgtGtCmdSE4-ow"];
+    
+    NSData *data = [[NSData alloc] initWithContentsOfURL:url];
```
boundsj: I understand that minimizing the networking related code here makes sense. However, since adding a blocking call on the main (UI) thread is not something you would normally want to do in an application, should we consider adding a note about this being used for brevity? I think this is the first occurrence of this API in this repo so there is no precedent, I think.

jmkiley: Agreed. All of our other examples do mention that you should use a background thread for GeoJSON,  so we should also add that.

friedbunny:
Rebased this branch onto `release-v3.5.0`, which now uses v3.5.0-beta.3.

@jmkiley @captainbarbosa you’ll need to pull these remote changes (probably via hard reset origin) and then re-run `./install.sh`.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,66 @@
...
+@interface ShapeCollectionFeatureExample () <MGLMapViewDelegate>
+@property (nonatomic) MGLMapView *mapView;
+@end
+
+@implementation ShapeCollectionFeatureExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:[MGLStyle lightStyleURLWithVersion:9]];
```
friedbunny: Let’s use an unversioned style URL here — this example doesn’t use RtS on any default map elements, so technically we don’t need to pin to a version.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,66 @@
...
+    self.mapView.delegate = self;
+    
+    [self.view addSubview:self.mapView];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    
+    // Parse the GeoJSON data.
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        NSURL *url = [NSURL URLWithString: @"https://api.mapbox.com/datasets/v1/mapbox/cj004g2ay04vj2xls3oqdu2ou/features?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpemc0YWlpNzAwcXUyd21ldDV6OWpxMGwifQ.A92RQZpwUgtGtCmdSE4-ow"];
```
friedbunny: Hardcoding an access token in an example makes me anxious. Can we replace this with an Xcode `<# token #>` that the user can then replace with their own?

friedbunny: [Like this.](https://github.com/mapbox/MapboxGeocoder.swift/blob/e9435cbc77727a8933f8f591105556b0a2453ea5/Geocoder%20Example/ViewController.m#L8)

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
1ec5: Here’s the guidance on using [`MGLStyleDefaultVersion`](https://www.mapbox.com/ios-sdk/api/3.4.2/Other%20Constants.html#/c:MGLStyle.h@MGLStyleDefaultVersion):

> The value of this constant may change in a future release of the SDK. If you use any feature that depends on a specific aspect of a default style – for instance, the minimum zoom level that includes roads – you may use the current value of this constant or the underlying style URL, but do not use the constant itself. Such details may change significantly from version to version.

1ec5: If we use `MGLStyleDefaultVersion` anywhere in our documentation, I’d accompany it with a comment explaining that we know what we’re doing – “don’t use this at home” if you’re using runtime styling.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
1ec5: Following up on https://github.com/mapbox/ios-sdk-examples/pull/69#discussion_r104875443, we should replace the Datasets API usage with local GeoJSON file usage. That way we can link to the GeoJSON file, allowing the developer to inspect its contents and compare its structure to the code here.

Otherwise, if we intend to use this opportunity to advertise the Datasets API, let’s provide a comment linking out to more information about it. This would be the first time many developers will’ve encountered that API. Opening a URL to a data set on the Mapbox account wouldn’t tell the developer anything about how they can create and edit their own data sets.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
jmkiley: I think that for now, keeping the version number is safest on our end, with a disclaimer like "Please see the [Styles API documentation](https://www.mapbox.com/api-documentation/#styles) or the current version number" either in the example or on the examples page.

In the future, as @friedbunny has suggested elsewhere, it would be worth looking at how to automate version number updates within examples.

Examples/Swift/ShapeCollectionFeatureExample.swift
```diff
@@ -0,0 +1,70 @@
...
+        
+        mapView.delegate = self
+        view.addSubview(mapView)
+    }
+    
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        
+        // Parse the GeoJSON data.
+        DispatchQueue.global().async {
+            let url = URL(string: "https://api.mapbox.com/datasets/v1/mapbox/cj004g2ay04vj2xls3oqdu2ou/features?access_token=\(MapboxAccessToken)")
```
friedbunny: This dataset doesn’t work with a non-`mapbox` access token:

```
fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSCocoaErrorDomain Code=256 "The file “features” couldn’t be opened." UserInfo={NSURL=https://api.mapbox.com/datasets/v1/mapbox/cj004g2ay04vj2xls3oqdu2ou/features?access_token=someactualaccesstoken
```

Examples/Swift/ShapeCollectionFeatureExample.swift
```diff
@@ -0,0 +1,70 @@
+import Mapbox;
+
+let MapboxAccessToken = "<# Access Token #>"
```
friedbunny: Sorry, but now that I see this external variable up here I’m thinking that it would be better to just include `<# Mapbox Access Token #>` in the URL string itself. In our Swift library examples the external variable exists so that we might throw an exception when the user hasn’t set it, which doesn’t need to be the case here.

Examples/Swift/ShapeCollectionFeatureExample.swift
jmkiley: Okay! I'll switch it to a geojson file :)

friedbunny: This dataset probably could have been made public, instead:

screen_shot_2017-03-14_at_8_35_34_pm

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
friedbunny: We have several geojson examples, but none/few showing the datasets API. @jmkiley has already moved to switch to geojson, but I think we should have some example (at some point) that moves beyond a flat file.

Examples/Swift/ShapeCollectionFeatureExample.swift
1ec5: Ah, I hadn’t considered making the data set public. That would’ve given us a URL we could link to in a comment.

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
1ec5: → #72

Examples/Swift/ShapeCollectionFeatureExample.swift
jmkiley: Agreed. It would be great to show how you can use the Datasets editor and Datasets API with the iOS SDK. I think that can be a priority once we have more DDS examples!

Examples/ObjectiveC/ShapeCollectionFeatureExample.m
```diff
@@ -0,0 +1,68 @@
+#import "ShapeCollectionFeatureExample.h"
+@import Mapbox;
+
+NSString *const MBXExampleShapeCollectionFeature = @"ShapeCollectionFeatureExample";
+NSString *const MapboxAccessToken = @"<#Access Token#>";
```
friedbunny: This is unused now.

Examples/Swift/ShapeCollectionFeatureExample.swift
```diff
@@ -0,0 +1,68 @@
+import Mapbox;
+
+@objc(ShapeCollectionFeatureExample_Swift)
+
+class ShapeCollectionFeatureExample_Swift: UIViewController, MGLMapViewDelegate {
+    
+    var mapView : MGLMapView!
```
friedbunny: Unnecessary colon space.

jmkiley:
Changed base branch to `master`

Examples.xcodeproj/project.pbxproj.orig
```diff
@@ -0,0 +1,1196 @@
+// !$*UTF8*$!
```
friedbunny: This looks like the product of a resolved merge conflict — should be safe to delete this file.

PR #70, opened 2017-03-11
Add DDS circle layer example

captainbarbosa:
Add data driven styling example using `MGLCircleStyleLayer`.

screen shot 2017-03-10 at 6 56 15 pm

cc @jmkiley

### DIFF
--- a/Examples/ObjectiveC/DDSCircleLayerExample.m
+++ b/Examples/ObjectiveC/DDSCircleLayerExample.m
@@ -1,0 +1,65 @@
+#import "DDSCircleLayerExample.h"
+@import Mapbox;
+
+NSString *const MBXExampleDDSCircleLayer = @"DDSCircleLayerExample";
+
+@interface DDSCircleLayerExample () <MGLMapViewDelegate>
+
+@property (nonatomic) MGLMapView *mapView;
+
+@end
+
+@implementation DDSCircleLayerExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    // Create a new map view using the Mapbox Light style.
+    self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds
+        styleURL:[MGLStyle lightStyleURLWithVersion:9]];
+    
+    self.mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    
+    // Set the map’s center coordinate and zoom level.
+    self.mapView.centerCoordinate = CLLocationCoordinate2DMake(38.897,-77.039);
+    self.mapView.zoomLevel = 10.5;
+    
+    self.mapView.delegate = self;
+    [self.view addSubview: self.mapView];
+}
+
+// Wait until the style is loaded before modifying the map style.
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    
+    // "mapbox://examples.2uf7qges" is a map ID referencing a tileset. For more
+    // more information, see mapbox.com/help/define-map-id/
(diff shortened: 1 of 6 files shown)


Examples/ObjectiveC/DDSCircleLayerExample.m
```diff
@@ -0,0 +1,62 @@
...
+    MGLSource *source = [[MGLVectorSource alloc] initWithIdentifier:@"trees" configurationURL:[NSURL URLWithString:@"mapbox://examples.2uf7qges"]];
+    
+    [self.mapView.style addSource:source];
+    
+    MGLCircleStyleLayer *layer = [[MGLCircleStyleLayer alloc] initWithIdentifier:[NSString stringWithFormat:@"tree-style"] source:source];
+    
+    // The source name from the source's TileJSON metadata: mapbox.com/api-documentation/#retrieve-tilejson-metadata
+    layer.sourceLayerIdentifier = @"yoshino-trees-a0puw5";
+    
+    // Categorical stops based on age of tree in years.
```
jmkiley: I don't think that this would be a categorical stop, since you're using `MGLInterpolationModeInterval`. Maybe you could say "stops" or "stops dictionary?"

Examples/ObjectiveC/DDSCircleLayerExample.m
```diff
@@ -0,0 +1,62 @@
...
+    [self.view addSubview: self.mapView];
+}
+
+// Wait until the style is loaded before modifying the map style.
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    MGLSource *source = [[MGLVectorSource alloc] initWithIdentifier:@"trees" configurationURL:[NSURL URLWithString:@"mapbox://examples.2uf7qges"]];
+    
+    [self.mapView.style addSource:source];
+    
+    MGLCircleStyleLayer *layer = [[MGLCircleStyleLayer alloc] initWithIdentifier:[NSString stringWithFormat:@"tree-style"] source:source];
```
boundsj: nit: It looks like "tree-style" could just be a simple NSString constant here (i.e. `@"tree-style"` since we don't actually make use of the string format API)

jmkiley:
👍 from me

Examples/ObjectiveC/DDSCircleLayerExample.m
```diff
@@ -0,0 +1,62 @@
...
+    self.mapView.centerCoordinate = CLLocationCoordinate2DMake(38.897,-77.039);
+    self.mapView.zoomLevel = 10.5;
+    
+    self.mapView.delegate = self;
+    [self.view addSubview: self.mapView];
+}
+
+// Wait until the style is loaded before modifying the map style.
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    MGLSource *source = [[MGLVectorSource alloc] initWithIdentifier:@"trees" configurationURL:[NSURL URLWithString:@"mapbox://examples.2uf7qges"]];
```
friedbunny: I think we should add a note here about what this source is. Mainly: what is `mapbox://examples.2uf7qges`, why isn’t it a style, how do I get this on my own?

Examples/Swift/DDSCircleLayerExample.swift
```diff
@@ -0,0 +1,56 @@
...
+        mapView.zoomLevel = 10.5
+
+        mapView.delegate = self
+        view.addSubview(mapView)
+    }
+    
+    // Wait until the style is loaded before modifying the map style.
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        
+        let source = MGLVectorSource(identifier: "population", configurationURL: URL(string: "mapbox://examples.2uf7qges")!)
```
friedbunny: This `identifier` doesn’t match the objc example.

captainbarbosa:
Woop! 🎉 🌈 🚀

PR #76, opened 2017-03-22
Set tint color to match map style

friedbunny:
Updates some recent examples’ tint colors to match their map styles. The main (and only) difference is that this colors the ℹ️ icon.

simulator screen shot mar 21 2017 8 47 07 pm

/cc @captainbarbosa


friedbunny:
This does technically call for updates to the ios-sdk screenshots for these examples, but that’s not terribly pressing.

PR #77, opened 2017-03-22
Switch to CocoaPods and automanaged signing

friedbunny:
This changes how the SDK is installed and updated. `install.sh` is dead (😐) — now it’s the more usual `pod install`.

A bit of preparation for running tests on CI.

/cc @mapbox/ios


friedbunny:
This repo now has CI builds on Bitrise: https://www.bitrise.io/app/9a144f2169b7c9e3 🌠

This workflow runs the two test targets, of which only `ExamplesUITests` has content — a single test called `testEveryExample` that iterates over the table view and taps each example, waits for it to load, and then goes back to the table view. 🙃

PR #83, opened 2017-04-19
Bumped podfile for v3.5.2

jmkiley:
cc @friedbunny


friedbunny:
Thanks! Also be sure to check-in changes to `Podfile.lock`.

jmkiley:
Added! Sorry about that :) @friedbunny

PR #86, opened 2017-05-12
[do not merge] Extrusions example

jmkiley:
Basic 3D extrusions example for when 3.6.0 lands.

screenshot 2017-05-12 10 46 22

cc @incanus


Examples/Swift/ExtrusionsExample.swift
```diff
@@ -0,0 +1,50 @@
...
+        // Set the map view camera's pitch and distance.
+        mapView.camera = MGLMapCamera(lookingAtCenter: mapView.centerCoordinate, fromDistance: 600, pitch: 60, heading: 0)
+        mapView.delegate = self
+        
+        view.addSubview(mapView)
+    }
+    
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        
+        // Access the Mapbox Streets source and use it to create a `MGLFillExtrusionLayer`.
```
friedbunny: `MGLFillExtrusionStyleLayer` seems to be the name of the layer class.

Examples/Swift/ExtrusionsExample.swift
```diff
@@ -0,0 +1,50 @@
...
+        let source = style.source(withIdentifier: "composite")
```
friedbunny: Is `"composite"` chosen for a particular reason (e.g., does this exist in the source) or is this an arbitrary name?

Examples/Swift/ExtrusionsExample.swift
```diff
@@ -0,0 +1,50 @@
...
+        let layer = MGLFillExtrusionStyleLayer(identifier: "buildings", source: source!)
```
friedbunny: Is it safe to force unwrap `source`?

Examples/Swift/ExtrusionsExample.swift
```diff
@@ -0,0 +1,50 @@
...
+        layer.sourceLayerIdentifier = "building"
+        
+        // Filter out buildings that should not extrude.
+        layer.predicate = NSPredicate(format: "extrude != false AND height >=0")
```
friedbunny: For my own curiosity — what constitutes a building with `extrude == false`? How is that determined?

Examples/Swift/ExtrusionsExample.swift
```diff
@@ -0,0 +1,50 @@
...
+        
+        // Set the fill extrusion height to the value for the building height attribute.
+        layer.fillExtrusionHeight = MGLStyleValue(interpolationMode: .identity, sourceStops: nil, attributeName: "height", options: nil)
+        layer.fillExtrusionOpacity = MGLStyleValue(rawValue: 0.75)
+        layer.fillExtrusionColor = MGLStyleValue(rawValue: .white)
+        
+        // Insert the fill extrusion layer below a POI label layer.
+        let symbolLayer = style.layer(withIdentifier: "poi-scalerank3")
+        style.insertLayer(layer, below: symbolLayer!)
```
friedbunny: Is `symbolLayer` safe to force unwrap? When dealing with layer identifiers, I think it’s going to be important to note that these are only guaranteed to exist in certain Mapbox sources/styles, and then only for specific versions.

/cc @1ec5

friedbunny:
Nice example. Since it requires v3.6.0, we should create a release branch here and retarget this PR against that.

Podfile
```diff
@@ -3,7 +3,7 @@ use_frameworks!
 
 target 'Examples' do
   # Pods for Examples
-  pod 'Mapbox-iOS-SDK-symbols', :podspec => 'https://raw.githubusercontent.com/mapbox/mapbox-gl-native/ios-v3.5.3/platform/ios/Mapbox-iOS-SDK-symbols.podspec'
+  pod 'Mapbox-iOS-SDK-symbols', :podspec => 'https://raw.githubusercontent.com/mapbox/mapbox-gl-native/ios-v3.6.0-alpha.1/platform/ios/Mapbox-iOS-SDK-symbols.podspec'
```
friedbunny: Let’s break the version bump out into its own commit.

jmkiley: Created a `release-v3.6.0` branch. Will update the podfile in there as needed!

Examples/Swift/ExtrusionsExample.swift
jmkiley: `"composite"` is the identifier of the source in the style. I'll clarify that in the comment!

Examples/Swift/ExtrusionsExample.swift
jmkiley: From what I've read [here](https://www.mapbox.com/vector-tiles/mapbox-streets-v7/), `extrude` may be `false` if it is a complex building that requires a lot of building parts to render it correctly (like the Colosseum). This tells the renderer to skip rendering the overall building, and instead render the individual building parts. 

This is mostly to catch some strange rendering that @incanus and I noticed when we were testing this out on Lower Manhattan. (which is a bug that he reported)

screen shot 2017-05-10 at 11 41 44 am

Examples/Swift/ExtrusionsExample.swift
jmkiley: I'll err on the side of safety 🙇‍♀️

Examples/Swift/ExtrusionsExample.swift
1ec5: According to the [Mapbox Streets source documentation](https://www.mapbox.com/vector-tiles/mapbox-streets-v7/#building):

> The `extrude` field is `true` or `false` depending one whether the object should be included in 3D-extrusion renderings. For example a complex building might have various `building:part` objects mapped with different heights, in addition to a building object representing the footprint of the entire building. Only the `building:part` objects are needed for 3D rendering, so the full footprint outline will have an `extrude` value of `false`.

For a lower-level understanding of this value, see the [OpenStreetMap documentation](https://wiki.openstreetmap.org/wiki/Simple_3D_buildings#Building_outline) about 3D building outlines.

Examples/Swift/ExtrusionsExample.swift
1ec5: See [this StackOverflow post](http://stackoverflow.com/a/43876838/4585461) for a more robust solution.

Examples/Swift/ExtrusionsExample.swift
1ec5: For a more robust solution (which may be overkill for this particular example), see [this extension method](https://github.com/mapbox/mapbox-navigation-ios/blob/20c082eea7946bb35b5895f8ca525275ce0bdc7f/MapboxNavigation/MGLVectorSource.swift#L5-L10) used in the navigation SDK. You’d filter `style.sources` by `isMapboxStreets` and [create a Streets source](https://github.com/mapbox/mapbox-navigation-ios/blob/238e3f2a52decba5489a0b44872bb6168fd59115/MapboxNavigation/RouteMapViewController.swift#L332-L343) if there isn’t one.

Examples/ObjectiveC/ExtrusionsExample.m
```diff
@@ -0,0 +1,56 @@
...
+
+@implementation ExtrusionsExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:[MGLStyle lightStyleURLWithVersion:9]];
+    
+    // Center the map on the Colosseum in Rome, Italy.
+    [mapView setCenterCoordinate:CLLocationCoordinate2DMake(41.8902, 12.4922)];
```
1ec5: Nit: use dot notation: `mapView.centerCoordinate = …`.

1ec5: Actually, we’re setting the camera right below. Let’s combine the two statements.

Examples/ObjectiveC/ExtrusionsExample.m
```diff
@@ -0,0 +1,56 @@
...
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    
+    // Access the Mapbox Streets source and use it to create a `MGLFillExtrusionStyleLayer`. The source identifier is `composite`. Use the `sources` property on a style to verify source identifiers.
+    MGLSource *source = [style sourceWithIdentifier:@"composite"];
+    MGLFillExtrusionStyleLayer *layer = [[MGLFillExtrusionStyleLayer alloc] initWithIdentifier:@"buildings" source:source];
+    layer.sourceLayerIdentifier = @"building";
+    
+    // Filter out buildings that should not extrude.
+    layer.predicate = [NSPredicate predicateWithFormat:@"extrude != false AND height >=0"];
```
1ec5: `extrude` can only be true or false, right? `extrude = true` reads better in this case.

Examples/ObjectiveC/ExtrusionsExample.m
```diff
@@ -0,0 +1,56 @@
...
+    
+    // Set the fill extrusion height to the value for the building height attribute.
+    layer.fillExtrusionHeight = [MGLStyleValue valueWithInterpolationMode:MGLInterpolationModeIdentity sourceStops:nil attributeName:@"height" options:nil];
```
1ec5: We should set `fillExtrusionBase` too, so that roofs and skywalks don’t fill the space below. The coliseum will probably look a little better that way, too.

Examples/Swift/ExtrusionsExample.swift
```diff
@@ -0,0 +1,53 @@
...
+    
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        
+        // Access the Mapbox Streets source and use it to create a `MGLFillExtrusionStyleLayer`. The source identifier is `composite`. Use the `sources` property on a style to verify source identifiers.
+        if let source = style.source(withIdentifier: "composite") {
+            let layer = MGLFillExtrusionStyleLayer(identifier: "buildings", source: source)
+            layer.sourceLayerIdentifier = "building"
+            
+            // Filter out buildings that should not extrude.
+            layer.predicate = NSPredicate(format: "extrude != false AND height >=0")
```
1ec5: <sub><sup>NIT</sup></sub>: missing a space between `>=` and `0`.

Examples/Swift/ExtrusionsExample.swift
```diff
@@ -0,0 +1,53 @@
...
+            
+            // Set the fill extrusion height to the value for the building height attribute.
+            layer.fillExtrusionHeight = MGLStyleValue(interpolationMode: .identity, sourceStops: nil, attributeName: "height", options: nil)
+            layer.fillExtrusionOpacity = MGLStyleValue(rawValue: 0.75)
+            layer.fillExtrusionColor = MGLStyleValue(rawValue: .white)
+            
+            // Insert the fill extrusion layer below a POI label layer. Use the `layers` property on a style to verify built-in layer identifiers.
```
1ec5: This comment may lead a developer to think they have to use the `layers` property to obtain the correct layer at runtime, whereas we merely intend to point out that it’s a helpful tool for discovering the layer identifier.

> If you aren’t sure what the layer is called, you can view the style in Mapbox Studio or iterate over the style’s `layers` property, printing out each layer’s `identifier`.

Examples/ObjectiveC/ExtrusionsExample.m
```diff
@@ -0,0 +1,56 @@
...
+    // Set the map view camera's pitch and distance.
+    mapView.camera = [MGLMapCamera cameraLookingAtCenterCoordinate:mapView.centerCoordinate fromDistance:600 pitch:60 heading:0];
+    mapView.delegate = self;
+    
+    [self.view addSubview:mapView];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    
+    // Access the Mapbox Streets source and use it to create a `MGLFillExtrusionStyleLayer`. The source identifier is `composite`. Use the `sources` property on a style to verify source identifiers.
```
captainbarbosa: It might be a good idea to reference the Style Specification here: https://www.mapbox.com/mapbox-gl-js/style-spec

1ec5: This comment refers to properties in the runtime styling API. There are so many differences between that API and the style specification’s documentation on [sources](https://www.mapbox.com/mapbox-gl-js/style-spec/#root-sources) that we’d want to link to the more detailed [API reference](https://www.mapbox.com/ios-sdk/api/3.5.4/Classes/MGLStyle.html) instead.

In any case, I think https://github.com/mapbox/ios-sdk-examples/pull/86#discussion_r116355291 is relevant here as well:

> If you aren’t sure what the `source` is called, you can view the style in Mapbox Studio or iterate over the style’s `sources` property, printing out each source’s identifier.

Examples/ObjectiveC/ExtrusionsExample.m
```diff
@@ -41,14 +40,14 @@ - (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
     layer.sourceLayerIdentifier = @"building";
     
     // Filter out buildings that should not extrude.
-    layer.predicate = [NSPredicate predicateWithFormat:@"extrude != false AND height >=0"];
```
1ec5: We _do_ want to extrude. https://github.com/mapbox/ios-sdk-examples/pull/86/#discussion_r116355227 was a suggestion to change this to `extrude = true`, not to remove the condition. By removing the condition, buildings like the Eiffel Tower will get extruded quite humorously.

Examples/ObjectiveC/ExtrusionsExample.m
```diff
@@ -0,0 +1,55 @@
...
+    // Access the Mapbox Streets source and use it to create a `MGLFillExtrusionStyleLayer`. The source identifier is `composite`. Use the `sources` property on a style to verify source identifiers.
+    MGLSource *source = [style sourceWithIdentifier:@"composite"];
+    MGLFillExtrusionStyleLayer *layer = [[MGLFillExtrusionStyleLayer alloc] initWithIdentifier:@"buildings" source:source];
+    layer.sourceLayerIdentifier = @"building";
+    
+    // Filter out buildings that should not extrude.
+    layer.predicate = [NSPredicate predicateWithFormat:@"height >= 0"];
+    
+    // Set the fill extrusion height to the value for the building height attribute.
+    layer.fillExtrusionHeight = [MGLStyleValue valueWithInterpolationMode:MGLInterpolationModeIdentity sourceStops:nil attributeName:@"height" options:nil];
```
1ec5: Remember to set `fillExtrusionBase` too: https://github.com/mapbox/ios-sdk-examples/pull/86#discussion_r116355240.

Examples/ObjectiveC/ExtrusionsExample.m
jmkiley: Awesome! I just edited it. Was running into issues because of https://github.com/mapbox/mapbox-gl-native/issues/9056

Examples/ObjectiveC/ExtrusionsExample.m
```diff
@@ -40,10 +40,11 @@ - (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
     layer.sourceLayerIdentifier = @"building";
     
     // Filter out buildings that should not extrude.
-    layer.predicate = [NSPredicate predicateWithFormat:@"height >= 0"];
+    layer.predicate = [NSPredicate predicateWithFormat:@"extrude == %@ AND height >= 0", @"true"];
```
1ec5: Nit: Per https://github.com/mapbox/mapbox-gl-native/issues/9056#issuecomment-302822052, this can be written `extrude == 'true'`.

PR #87, opened 2017-05-12
[WIP] Heatmap example

jmkiley:
Ported [this example](https://www.mapbox.com/mapbox-gl-js/example/heatmap/) from GL-JS. I switched the data source to use the [USGS earthquake live data](https://earthquake.usgs.gov/earthquakes/feed/v1.0/geojson.php).

- [X] Write example
- [x] Add comments
- [ ] Incorporate revisions

screenshot 2017-05-12 11 30 13


Examples/Swift/PointHeatmapExample.swift
```diff
@@ -0,0 +1,54 @@
...
+        mapView.delegate = self
+        
+        view.addSubview(mapView)
+    }
+    
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        
+        // Parse GeoJSON data from USGS on earthquakes in the past week.
+        DispatchQueue.global().async {
+            guard let url = URL(string: "https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_week.geojson") else { return }
```
friedbunny: This URL should never fail to initialize and can be safely force-unwrapped.

Examples/Swift/PointHeatmapExample.swift
```diff
@@ -0,0 +1,54 @@
...
+            DispatchQueue.main.async {
+                self.displayEarthquakes(url: url, style: style)
+            }
```
friedbunny: Can you say more about what the intention is here, re: threading? It appears that you’re creating the URL on a background thread, then doing the actual fetching and rendering on the main thread.

Examples/Swift/PointHeatmapExample.swift
```diff
@@ -0,0 +1,54 @@
...
+    func displayEarthquakes(url: URL, style: MGLStyle) {
+        let symbolSource = MGLSource(identifier: "symbol-source")
+        let symbolLayer = MGLSymbolStyleLayer(identifier: "place-city-sm", source: symbolSource)
+        
+        // Set the MGLShapeSourceOptions to allow clustering.
+        let options = [MGLShapeSourceOption.clustered: true,
+                       MGLShapeSourceOption.clusterRadius: 20,
+                       MGLShapeSourceOption.maximumZoomLevel: 15] as [MGLShapeSourceOption : Any]
+        
+        let earthquakeSource = MGLShapeSource(identifier: "earthquakes", url: url, options: options)
```
friedbunny: If this is where the remote fetching happens, should there be some protection against it failing?

Examples/Swift/PointHeatmapExample.swift
friedbunny: The URLs themselves are also different — the objc example uses the static GL-JS version of this feed.

jmkiley:
screenshot 2017-05-23 10 20 09

Changed the opacity and colors, switched to dark style.

Examples/Swift/PointHeatmapExample.swift
jmkiley: I changed the source to a Mapbox hosted one. Will look into using this live data source for a tutorial/different example.

jmkiley:
Is there anything else to address with this? @friedbunny @captainbarbosa

captainbarbosa:
No, I don't think so @jmkiley - looks good! 👍

PR #90, opened 2017-06-01
Add fastlane TestFlight infrastructure

friedbunny:
Setting up a Bitrise workflow that deploys to TestFlight on every commit to `master`.

/cc @jmkiley @captainbarbosa


friedbunny:
Ended up storing the credentials and such entirely in Bitrise, so the changes in this repo are minimal.

PR #94, opened 2017-06-21
[WIP] Fill pattern example

captainbarbosa:
Working on a new example that demonstrates how to add a fill pattern within a [`MGLFillStyleLayer`](MGLFillStyleLayer).


captainbarbosa:
Here's what its looking like so far. The marked region represents the [FCC's no drone zone](https://www.faa.gov/uas/where_to_fly/no_drone_zone/) around Washington, D.C.

simulator screen shot jun 26 2017 12 54 01 pm

Examples/Assets.xcassets/pattern.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "pattern.pdf",
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "2x"
```
friedbunny: Select “Scales = Single Scale” for vector assets.

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,86 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        let mapView = MGLMapView(frame: view.bounds)
+        
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.styleURL = MGLStyle.darkStyleURL(withVersion: MGLStyleDefaultVersion)
+        
+        // Set the map’s center coordinate and zoom level
+        mapView.setCenter(CLLocationCoordinate2D(latitude: 38.849534447295895, longitude: -77.03922271728514), zoomLevel: 8.5, animated: false)
```
friedbunny: Nit: use less precise numbers for the coordinate.

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,86 @@
+import Mapbox
+
+@objc(FillPatternExample_Swift)
+
+class FillPattern_Swift: UIViewController, MGLMapViewDelegate {
+    var fillPatternImage: UIImage!
```
friedbunny: Because this is only used in `didFinishLoadingStyle`, this doesn’t have to be a property of the view controller. Instead, we can do all of the setup in `didFinishLoadingStyle`.

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,86 @@
...
+        
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.styleURL = MGLStyle.darkStyleURL(withVersion: MGLStyleDefaultVersion)
+        
+        // Set the map’s center coordinate and zoom level
+        mapView.setCenter(CLLocationCoordinate2D(latitude: 38.849534447295895, longitude: -77.03922271728514), zoomLevel: 8.5, animated: false)
+        view.addSubview(mapView)
+        
+        // Set the UIImage to be used for the fill pattern
+        fillPatternImage = UIImage(named: "pattern")
```
friedbunny: Have you considered drawing this pattern in code using Core Graphics?

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,86 @@
...
+        
+        mapView.delegate = self
+    }
+    
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        
+        // Add the fill pattern image to used by the style layer
+        style.setImage(fillPatternImage.withRenderingMode(.alwaysTemplate), forName: "pattern")
+        
+        let coordinates = [
```
friedbunny: This is fine, but it could be less verbose (in Swift, anyway) if the array were structured [like this one](https://github.com/mapbox/ios-sdk-examples/blob/dcb2167e60e356b3179a5a13d6799377d995933f/Examples/Swift/RuntimeAnimateLineExample.swift#L85-L161).

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,86 @@
...
+        // Create a shape source based off of a polygon object
+        let shape = MGLPolygon(coordinates: coordinates, count: UInt(coordinates.count))
+        let source = MGLShapeSource(identifier: "circle", shape: shape, options: nil)
+        style.addSource(source)
+        
+        // Create a style layer to be used with the shape source
+        let layer = MGLFillStyleLayer(identifier: "circle", source: source)
+        
+        // Set the fill pattern and opacity for the style layer
+        layer.fillPattern = MGLStyleValue<NSString>(rawValue: "pattern")
```
friedbunny: This seems a little magical — perhaps a more specific name than `pattern` and a bit more detailed of an explanation would help.

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,86 @@
...
+class FillPattern_Swift: UIViewController, MGLMapViewDelegate {
+    var fillPatternImage: UIImage!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        let mapView = MGLMapView(frame: view.bounds)
+        
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.styleURL = MGLStyle.darkStyleURL(withVersion: MGLStyleDefaultVersion)
```
friedbunny: This example uses a named layer in the dark style, so it should be pinned to the specific version of that style and not the generic default style version (which can be updated with any release).

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,86 @@
...
+        // Create a style layer to be used with the shape source
+        let layer = MGLFillStyleLayer(identifier: "circle", source: source)
+        
+        // Set the fill pattern and opacity for the style layer
+        layer.fillPattern = MGLStyleValue<NSString>(rawValue: "pattern")
+        layer.fillOpacity = MGLStyleValue(rawValue: 0.5)
+        
+        // Place the style layer below the layer contining city labels, if the
+        // layer is not found, add the style layer above all layers
+        if let cityLabels = style.layer(withIdentifier: "place-city-sm") {
```
friedbunny: We should note that this is style-specific and may not exist in custom styles.

Examples/Swift/SimpleMapViewExample.swift
```diff
@@ -5,12 +5,13 @@ import Mapbox
 class SimpleMapViewExample_Swift: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
-        
+        self.navigationController?.isNavigationBarHidden = true
```
friedbunny: These changes look errant?

captainbarbosa: Yep 🙈  fixed - was debugging an unrelated issue

Examples.xcodeproj/project.pbxproj
```diff
@@ -350,9 +355,12 @@
 				968247191C5C115000494AB8 /* DrawingAGeoJSONLineExample.m */,
 				9682470E1C5BF12700494AB8 /* DrawingAMarkerExample.m */,
 				9682471F1C5C1C0400494AB8 /* DrawingAPolygonExample.m */,
+				9682471F1C5C1C0400494AB8 /* DrawingAPolygonExample.m */,
 				3E085B821EC526E500163C99 /* ExtrusionsExample.m */,
 				3E3FB66D1F0588B3004512C6 /* LightExample.m */,
 				3E3FB66F1F0589AD004512C6 /* PointHotspotExample.m */,
+				055ABCC01EFB139C0063BACA /* FillPatternExample.m */,
+				968247291C5C1FF800494AB8 /* PointConversionExam
```
friedbunny: This seems bork.

Examples.xcodeproj/project.pbxproj
```diff
@@ -350,9 +355,12 @@
 				968247191C5C115000494AB8 /* DrawingAGeoJSONLineExample.m */,
 				9682470E1C5BF12700494AB8 /* DrawingAMarkerExample.m */,
 				9682471F1C5C1C0400494AB8 /* DrawingAPolygonExample.m */,
+				9682471F1C5C1C0400494AB8 /* DrawingAPolygonExample.m */,
```
friedbunny: Dupe — this is usually harmless, but unnecessary nonetheless.

Examples/ObjectiveC/FillPatternExample.m
```diff
@@ -0,0 +1,69 @@
...
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    
+    // Set the UIImage to be used for the fill pattern.
+    UIImage *fillPatternImage = [UIImage imageNamed:@"stripe-pattern"];
+    
+    // Add the fill pattern image to used by the style layer.
+    [style setImage:fillPatternImage forName:@"stripe-pattern"];
+    
+    // "mapbox://examples.0cd7imtl" is a map ID referencing a tileset. For more
```
friedbunny: > is a map ID referencing a tileset.

That’s what this is technically, but what content does it contain?

Examples/ObjectiveC/FillPatternExample.m
```diff
@@ -0,0 +1,69 @@
...
+    // more information, see mapbox.com/help/define-map-id/
+    MGLSource *source = [[MGLVectorSource alloc] initWithIdentifier:@"drone-restrictions" configurationURL:[NSURL URLWithString:@"mapbox://examples.0cd7imtl"]];
+    
+    [style addSource:source];
+    
+    // Create a style layer to be used with the vector source.
```
friedbunny: > to be used

Would “using” be more precise? We’ve already init’d the source on L38 and are adding it to the layer here.

Examples/ObjectiveC/FillPatternExample.m
```diff
@@ -0,0 +1,69 @@
...
+    MGLFillStyleLayer *layer = [[MGLFillStyleLayer alloc] initWithIdentifier:@"drone-restrictions-style" source:source];
+    
+    // Set the source's identifier using the source name its
+    // TileJSON metadata: mapbox.com/api-documentation/#retrieve-tilejson-metadata
```
friedbunny: This comment seems incomplete.

Examples/ObjectiveC/FillPatternExample.m
```diff
@@ -0,0 +1,69 @@
...
+    layer.fillPattern = [MGLStyleValue valueWithRawValue:@"stripe-pattern"];
+    layer.fillOpacity = [MGLStyleValue valueWithRawValue:@0.5];
+    
+    // Insert the pattern style layer below the layer contining city labels. If the
+    // layer is not found, the style layer will be added above all other layers within the
+    // Mapbox Dark style. NOTE: The "place-city-sm" layer is specific to the Mapbox Dark style.
+    // Refer to the layers list in Mapbox Studio to confirm which layers are available for
+    // use when working with a custom style.
+    MGLStyleLayer *cityLabels = [style layerWithIdentifier:@"place-city-sm"];
+    if (cityLabels != nil) {
```
friedbunny: Checking for `!= nil` is necessary in Swift, but not Objective-C.

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,60 @@
...
+        view.addSubview(mapView)
+        
+        mapView.delegate = self
+    }
+    
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {
+        
+        // Set the UIImage to be used for the fill pattern.
+        var fillPatternImage: UIImage!
+        fillPatternImage = UIImage(named: "stripe-pattern")
```
friedbunny: Might as well combine these two lines in Swift, too.

Examples.xcodeproj/project.pbxproj
captainbarbosa: Donezo

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,59 @@
...
+        let layer = MGLFillStyleLayer(identifier: "drone-restrictions-style", source: source)
+        
+        // Set the source's identifier using the source name retrieved from its
+        // TileJSON metadata: mapbox.com/api-documentation/#retrieve-tilejson-metadata
+         layer.sourceLayerIdentifier = "drone-restrictions-3f6lsg"
+        
+        // Set the fill pattern and opacity for the style layer. The MGLStyleValue
+        // object is a generic container for a style attribute value. In this case,
+        // it is a reference to the fillPatternImage.
+        layer.fillPattern = MGLStyleValue<NSString>(rawValue: "stripe-pattern")
```
1ec5: Nit: remove `<NSString>`, since the compiler can infer it.

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,59 @@
...
+
+class FillPattern_Swift: UIViewController, MGLMapViewDelegate {
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Set the map’s size, style, center coordinate, and zoom level.
+        let mapView = MGLMapView(frame: view.bounds)
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.styleURL = MGLStyle.darkStyleURL(withVersion: 9)
```
1ec5: Is this example dependent on any implementation details in v9? If not and it’s purely additive, then we can use `MGLStyle.darkStyleURL()`.

Examples/Swift/FillPatternExample.swift
```diff
@@ -0,0 +1,59 @@
...
+        // "mapbox://examples.0cd7imtl" is a map ID referencing a tileset containing vector data.
+        // For more information, see mapbox.com/help/define-map-id/
+        let source = MGLVectorSource(identifier: "drone-restrictions", configurationURL: URL(string: "mapbox://examples.0cd7imtl")!)
+        style.addSource(source)
+        
+        // Create a style layer using the vector source.
+        let layer = MGLFillStyleLayer(identifier: "drone-restrictions-style", source: source)
+        
+        // Set the source's identifier using the source name retrieved from its
+        // TileJSON metadata: mapbox.com/api-documentation/#retrieve-tilejson-metadata
```
1ec5: Mapbox Studio provides a user-friendly way to see these layers without having to construct an API request by hand.

Examples/Swift/FillPatternExample.swift
captainbarbosa: Per @friedbunny's comment in https://github.com/mapbox/ios-sdk-examples/pull/94#discussion_r124073373, I originally had `MGLStyle.darkStyleURL()`, but it was suggested that we should pin the version to a specific number and explain that relying on simply the default version may become problematic if the current version differs significantly. Although, the `place-city-sm` label is an important layer that I feel wouldn't be drastically changed anytime soon, so I'm ok reverting it back to `MGLStyle.darkStyleURL()` , and adding more specific commentary instead.

1ec5: Ah, no, what you’ve got makes sense, given the reliance on `place-city-sm`.

captainbarbosa: Cool, I'll keep it then 💯

PR #95, opened 2017-06-22
Removed withVersion:9

jmkiley:
Removed `withVersion:` from the current examples since it is not necessary in v3.6.0.

### DIFF
--- a/Examples/ObjectiveC/ExtrusionsExample.m
+++ b/Examples/ObjectiveC/ExtrusionsExample.m
@@ -21,7 +21,7 @@ NSString *const MBXExample3DExtrusions = @"ExtrusionsExample";
 - (void)viewDidLoad {
     [super viewDidLoad];
     
-    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:[MGLStyle lightStyleURLWithVersion:9]];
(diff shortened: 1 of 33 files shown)


Examples/ObjectiveC/ExtrusionsExample.m
```diff
@@ -21,7 +21,7 @@ @implementation ExtrusionsExample
 - (void)viewDidLoad {
     [super viewDidLoad];
     
-    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:[MGLStyle lightStyleURLWithVersion:9]];
+    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:[MGLStyle lightStyleURL]];
```
1ec5: This example relies on details about the style’s existing source. We should keep specifying the version in such cases.

captainbarbosa:
Awesome @jmkiley, thanks for the heads up. I'm working on a fill pattern example so I'll be sure to follow the new format being used.

friedbunny:
@captainbarbosa if your fill pattern example relies on data-driven styling at all, it should stay with a versioned style URL.

jmkiley:
Awesome! I'm going to go back through tomorrow to check for any other potential source layer related issues. I'll add comments to make reasoning behind the choice more clear.

PR #96, opened 2017-06-23
[WIP] MGLLight Example

jmkiley:


jmkiley:
Current example:
screenshot 2017-06-26 11 31 22

jmkiley:
I think this is ready for review @friedbunny @fabian-guerra

Examples/ObjectiveC/LightExample.m
```diff
@@ -0,0 +1,70 @@
...
+
+NSString *const MBXExampleLight = @"LightExample";
+
+@implementation LightExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    // Set the map style to Mapbox Light Style version 9. The map's source will be queried later in this example.
+    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:[MGLStyle lightStyleURLWithVersion:9]];
```
fabian-guerra: Is it required version 9?

Examples/ObjectiveC/LightExample.m
```diff
@@ -0,0 +1,70 @@
...
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    
+    // Add a MGLFillExtrusionStyleLayer.
+    [self addFillExtrusionLayer:style];
+    
+    // Create an MGLLight object.
+    MGLLight *light = [[MGLLight alloc] init];
+    
+    // Create an MGLSphericalPosition and set the radial, azimuthal, and polar values.
```
friedbunny: For those who aren’t very mathematically inclined I’m not sure this sentence has much meaning. 😉

I see that the docs for [`MGLSphericalPosition`](https://www.mapbox.com/ios-sdk/api/3.6.0-rc.1/Structs/MGLSphericalPosition.html) have short explanations of these params, but perhaps we could also bring even more concise (one sentence each?) versions here.

Examples/ObjectiveC/LightExample.m
```diff
@@ -0,0 +1,70 @@
...
+    MGLSphericalPosition position = MGLSphericalPositionMake(5, 180, 80);
+    light.position = [MGLStyleValue valueWithRawValue:[NSValue valueWithMGLSphericalPosition:position]];
+    
+    // Set the light anchor to the map and add the light object to the map view's style.
```
friedbunny: > Set the light anchor to the map

Can we add more detail here — why do we do this and what’s the alternative?

Examples/ObjectiveC/LightExample.m
```diff
@@ -0,0 +1,70 @@
...
+                                                      sourceStops:nil
+                                                      attributeName:@"min_height"
+                                                      options:nil];
+    layer.fillExtrusionHeight = [MGLStyleValue valueWithInterpolationMode:MGLInterpolationModeIdentity
+                                                      sourceStops:nil
+                                                      attributeName:@"height"
+                                                      options:nil];
+    layer.fillExtrusionOpacity = [MGLStyleValue valueWithRawValue:@0.75];
+    layer.fillExtrusionColor = [MGLStyleValue valueWithRawValue:[UIColor whiteColor]];
+    MGLStyleLayer *symbolLayer = [style layerWithIdentifier:@"poi-scalerank3"];
```
friedbunny: It might be useful to point out that this layer name is specific to certain Mapbox styles.

friedbunny: This logic also differs from the Swift implementation, where it falls back to a plain `addLayer` if the label layer doesn’t exist.

Examples/ObjectiveC/LightExample.m
friedbunny: This is correct — Light v9 is the current version and this example uses style-specific runtime styling features, so it should be pinned.

Examples/Swift/LightExample.swift
```diff
@@ -0,0 +1,55 @@
...
+        
+        // Center the map on the Flatiron Building in New York, NY.
+        mapView.camera = MGLMapCamera(lookingAtCenter: CLLocationCoordinate2D(latitude: 40.7411, longitude: -73.9897), fromDistance: 600, pitch: 45, heading: 200)
+        mapView.tintColor = .gray
+        
+        view.addSubview(mapView)
+    }
+    
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle)
+    {
```
friedbunny: Nit: inconsistent backet placement.

Examples/ObjectiveC/LightExample.m
friedbunny: Though, using the Light style in an example that deals with a completely different type of “light” could be confusing.

friedbunny:
In general this example looks good, but I think the “light” concept could use more elucidation — e.g., it’s not clear from the screenshot or comments what this example is accomplishing.

- Could this have a toggle button or an animated/delayed transition?
- What do the light options and parameters do? How can we show/explain other possibilities?

jmkiley:
I originally had a slider that changed the radial position. Removed it to try to keep it simple, but you're right that showing the differences is helpful. Will bring that back 👍

jmkiley:
Simulator gif, but revised the example to include a slider for the polar value.
light

Examples/ObjectiveC/LightExample.m
```diff
@@ -0,0 +1,100 @@
...
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    // Set the map style to Mapbox Light Style version 9. The map's source will be queried later in this example.
+    self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:[MGLStyle streetsStyleURLWithVersion:9]];
+    self.mapView.delegate = self;
+    
+    // Center the map on the Flatiron Building in New York, NY.
+    self.mapView.camera = [MGLMapCamera cameraLookingAtCenterCoordinate:CLLocationCoordinate2DMake(40.7411, -73.9897) fromDistance:600 pitch:45 heading:200];
+    self.mapView.tintColor = [UIColor grayColor];
```
friedbunny: Thanks for setting a matching tint, but this is no longer matchy-matchy.

Examples/ObjectiveC/LightExample.m
```diff
@@ -0,0 +1,100 @@
...
+    // Add a MGLFillExtrusionStyleLayer.
+    [self addFillExtrusionLayer:style];
+    
+    // Create an MGLLight object.
+    self.light = [[MGLLight alloc] init];
+    
+    // Create an MGLSphericalPosition and set the radial, azimuthal, and polar values.
+    // Radial : Distance from the center of the base of an object to its light.
+    // Azimuthal : Position of the light relative to its anchor.
+    // Polar : The height of the light.
```
friedbunny: These are a big improvement! I realize it’d be essentially duplicating the API docs, but can we also say what unit each argument is in?

Examples/ObjectiveC/LightExample.m
```diff
@@ -0,0 +1,100 @@
...
+    MGLSphericalPosition position = MGLSphericalPositionMake(5, 180, 80);
+    self.light.position = [MGLStyleValue valueWithRawValue:[NSValue valueWithMGLSphericalPosition:position]];
+    
+    // Set the light anchor to the map and add the light object to the map view's style. The light anchor can be the viewport (or rotates with the viewport) or the map (rotates with the map).
+    // self.light.anchor = [MGLStyleValue valueWithRawValue:[NSValue valueWithMGLLightAnchor:MGLLightAnchorViewport]];
```
friedbunny: Ah, I mistook this for accidental leftover code — do you think `MGLLightAnchorViewport` could be worked into the comment instead?

Issue #98, opened 2017-06-29
Rename heat map to hotspot map


1ec5:
The “heat map” example added in #87 actually displays a hotspot map: mapbox/DEPRECATED-mapbox-gl#27. Heat maps are a separate feature that we’re working on and will want to [make a splash about](https://www.mapbox.com/blog/heatmaps-preview/) in the not-too-distant future. We should rename the example to avoid confusion.

/cc @jmkiley @friedbunny @captainbarbosa @mourner

jmkiley:
Addressed in https://github.com/mapbox/ios-sdk-examples/pull/97/

PR #97, opened 2017-06-29
[WIP] Update examples for 3.6.0

jmkiley:
This PR:
- [X] Adds a fill extrusion example
- [X] Adds a light example
- [X] Removes style version numbers from examples (as necessary)
- [x] Bumps podfile to v3.6.0 (Will do once 3.6.0 lands)

cc @fabian-guerra

PR #99, opened 2017-06-29
Renamed heatmap to hotspot

jmkiley:
Addresses https://github.com/mapbox/ios-sdk-examples/issues/98


jmkiley:
Addressed in https://github.com/mapbox/ios-sdk-examples/pull/97/

PR #108, opened 2017-08-16
Custom callout: stay open during panning (and some refactoring)

friedbunny:
Implements MGLCalloutView’s `anchoredToAnnotation` and `dismissesAutomatically` from https://github.com/mapbox/mapbox-gl-native/pull/7646 in our custom callout view example.

Because of https://github.com/mapbox/mapbox-gl-native/issues/9228, this also overrides the `center` property of the custom callout view, otherwise its position will shift undesirably when the user first interacts with the map.

Also:
- Selects the annotation in `viewDidLoad` so that the callout appears automatically.
- Moves the annotation property checks to `mapView:annotationCanShowCallout:`, which seems like a more natural fit than `mapView:calloutViewForAnnotation:`.
- Fixes a really annoying intermittent hairline gap between the arrow and label of the callout view. 😑
- Re-adds lazy vars, because [that Swift bug](https://bugs.swift.org/browse/SR-1825) has been fixed.
- Updates the project for CocoaPods 1.3.1.

/cc @jmkiley @frederoni @captainbarbosa


friedbunny:
👋 @jmkiley @frederoni

PR #116, opened 2017-09-20
Adopt Swift 4

### LINKED ISSUE #113: Support Swift 3.2/4.0
It’s time to move to Swift 3.2/4.0. It would be nice to avoid duplicating files for different versions, but how possible that will be remains to be seen.

- [Compatibility notes](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Compatibility.html)

/cc @jmkiley @captainbarbosa

friedbunny:
Fixes #113. Most of the changes here were automatic. The only significant code changes have to do with making `@objc` explicit.

Tests are failing right now because Bitrise is still on Xcode 8 — will flip that switch after #115 is merged.

/cc @jmkiley @captainbarbosa


friedbunny:
Updated Bitrise to use Xcode 9.

friedbunny:
Switched Bitrise to use the iOS 10 simulator — we should be sure to switch this back to the default once the iOS 11 OpenGL slowness is fixed by Apple.

friedbunny:
> Just curious, what is with the all the `@objc` additions?

Previous Swift versions were aggressive about adding Objective-C compatibility — this was implicit, but can also be expressed by adding `@objc` to the declaration. This came at the cost of performance and complexity.

Swift 4 infers `@objc` much less frequently and asks you to manually specify `@objc` for those methods that don’t absolutely require it — such as methods that you have chosen to refer to by selector (which is most of the ones in this PR).

The [Swift Evolution proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0160-objc-inference.md) for this has a lot of technical background and links to the discussion/implementation.

Issue #82, opened 2017-04-17
Animated polyline annotation


1ec5:
We should have an example of appending or otherwise mutating a polyline annotation on the fly, using `-[MGLPolyline appendCoordinates:count:]`. This would accomplish the same visual effect as “[Animated line](https://www.mapbox.com/ios-sdk/examples/runtime-animate-line/)” but demonstrate an easier and more performant approach than what the runtime styling API is currently capable of.

/cc @jmkiley @incanus

1ec5:
This is essentially how it works:

```swift
mapView.addAnnotation(polyline)
polyline.appendCoordinate(coordinate)
// there is no line 3 – the map updates automatically
```

PR #121, opened 2017-10-27
Add example for live data

jmkiley:
Ports https://www.mapbox.com/mapbox-gl-js/example/live-geojson/

- [x] Finish examples(https://developer.apple.com/documentation/dispatch/dispatchsourcetimer).
- [x] Add comments.


1ec5:
Given that the data source is in GeoJSON format, a shape source is definitely the right tool for the example use case we’re illustrating here.

Now we’re showing how to set shape sources dynamically here and in “[Animate a line](https://www.mapbox.com/ios-sdk/examples/animate-a-line/)”, we probably need to prioritize #82 for showing how to animate annotations, because the latter approach is more performant and cleaner (better for basic and heavy-duty use cases), even though it currently offers fewer styling options.

Examples/ObjectiveC/LiveDataExample.m
```diff
@@ -0,0 +1,51 @@
...
+    mapView.delegate = self;
+    
+    mapView.tintColor = [UIColor darkGrayColor];
+    [self.view addSubview:mapView];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    // Add a source to the map. https://wanderdrone.appspot.com/ generates coordinates for simulated paths.
+    NSURL *url = [NSURL URLWithString:@"https://wanderdrone.appspot.com/"];
+    _source = [[MGLShapeSource alloc] initWithIdentifier:@"drone-source" URL:url options:nil];
```
1ec5: The source identifier doesn’t need to contain “source”; it can be `wanderdrone` or simply `drones`.

Examples/ObjectiveC/LiveDataExample.m
```diff
@@ -0,0 +1,51 @@
...
+    droneLayer.iconImageName = [MGLStyleValue valueWithRawValue:@"rocket-15"];
+    [style addLayer:droneLayer];
+    
+    // Create a timer that calls the `updateUrl` function every 1.5 seconds.
+    _timer = [NSTimer scheduledTimerWithTimeInterval:1.5 target:self selector:@selector(updateURL) userInfo:nil repeats:YES];
+}
+
+- (void)updateURL {
+    // Update the icon's position by updating the `url` property on the source.
+    NSURL *url = [NSURL URLWithString:@"https://wanderdrone.appspot.com/"];
```
1ec5: The source starts out with this URL. To avoid repetition, we can pull the URL out of `_source` and set it on itself. That may look kind of strange, but I think the intent is clear.

jmkiley:
Thanks @1ec5! Addressed your comments. Agreed, finishing #82 is on my list after #124.

Examples/ObjectiveC/LiveDataExample.m
```diff
@@ -0,0 +1,50 @@
...
+@end
+
+@implementation LiveDataExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:[MGLStyle darkStyleURL]];
+    mapView.autoresizingMask = UIViewAutoresizingFlexibleHeight | UIViewAutoresizingFlexibleWidth;
+    mapView.delegate = self;
```
captainbarbosa: Can you add a comment above this line that [matches this one](https://github.com/mapbox/ios-sdk-examples/blob/master/Examples/Swift/DrawingACustomMarkerExample.swift#L21) for consistency across our examples?

Examples/ObjectiveC/LiveDataExample.m
```diff
@@ -0,0 +1,50 @@
...
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    // Add a source to the map. https://wanderdrone.appspot.com/ generates coordinates for simulated paths.
+    NSURL *url = [NSURL URLWithString:@"https://wanderdrone.appspot.com/"];
+    _source = [[MGLShapeSource alloc] initWithIdentifier:@"wanderdrone-source" URL:url options:nil];
+    [style addSource:_source];
+    
+    // Add an icon to the map to represent the drone's coordinate.
+    MGLSymbolStyleLayer *droneLayer = [[MGLSymbolStyleLayer alloc] initWithIdentifier:@"wanderdrone-layer" source:_source];
+    droneLayer.iconImageName = [MGLStyleValue valueWithRawValue:@"rocket-15"];
```
captainbarbosa: I would clarify where this symbol is coming from, since its confusing that we didn't include the asset for it in this example. In this case, the SDK can automatically reference Maki icons listed here: https://www.mapbox.com/maki-icons/

Examples/Swift/LiveDataExample.swift
```diff
@@ -0,0 +1,49 @@
...
+    
+    var source : MGLShapeSource!
+    var timer = Timer()
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        let mapView = MGLMapView(frame: view.bounds,
+                                 styleURL: MGLStyle.darkStyleURL(withVersion: 9))
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.delegate = self
```
captainbarbosa: Same as in the Objective-C example, we should add a comment above this line that [matches this one](https://github.com/mapbox/ios-sdk-examples/blob/master/Examples/Swift/DrawingACustomMarkerExample.swift#L21).

Examples/Swift/LiveDataExample.swift
```diff
@@ -0,0 +1,49 @@
...
+    func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle)
+    {
+        if let url = URL(string: "https://wanderdrone.appspot.com/") {
+            // Add a source to the map. https://wanderdrone.appspot.com/ generates coordinates for simulated paths.
+            source = MGLShapeSource(identifier: "wanderdrone", url: url, options: nil)
+            style.addSource(source)
+            
+            // Add an icon to the map to represent the drone's coordinate.
+            let droneLayer = MGLSymbolStyleLayer(identifier: "wanderdrone-layer", source: source)
+            droneLayer.iconImageName = MGLStyleValue(rawValue: "rocket-15")
```
captainbarbosa: Same as in the Objective-C example, lets clarify where that rocket icon is coming from.

jmkiley:
Thanks @captainbarbosa! Addressed your feedback.

Examples/ObjectiveC/LiveDataExample.m
1ec5: > In this case, the SDK can automatically reference Maki icons

But only because the initial style includes the Maki icons in its sprite sheet, right?

Examples/ObjectiveC/LiveDataExample.m
```diff
@@ -0,0 +1,53 @@
...
+    // Set the map view‘s delegate property.
+    mapView.delegate = self;
+    
+    [self.view addSubview:mapView];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    // Add a source to the map. https://wanderdrone.appspot.com/ generates coordinates for simulated paths.
+    NSURL *url = [NSURL URLWithString:@"https://wanderdrone.appspot.com/"];
+    _source = [[MGLShapeSource alloc] initWithIdentifier:@"wanderdrone-source" URL:url options:nil];
```
1ec5: There’s no need for `-source` in a source identifier and `-layer` in a layer identifier. Just thinking that developers might start [cargo-culting](http://www.catb.org/~esr/jargon/html/C/cargo-cult-programming.html) identifiers if they see this example.

Examples/Swift/LiveDataExample.swift
```diff
@@ -0,0 +1,51 @@
...
+            style.addLayer(droneLayer)
+            
+            // Create a timer that calls the `updateUrl` function every 1.5 seconds.
+            timer = Timer.scheduledTimer(timeInterval: 1.5, target: self, selector: #selector(updateUrl), userInfo: nil, repeats: true)
+        }
+    }
+    
+    @objc func updateUrl() {
+        // Update the icon's position by setting the `url` property on the source.
+            source.url = source.url
```
1ec5: Nit: too much indentation.

Examples/ObjectiveC/LiveDataExample.m
```diff
@@ -0,0 +1,53 @@
...
+    [style addSource:_source];
+    
+    // Add a Maki icon to the map to represent the drone's coordinate. For more information about Maki icons, see https://www.mapbox.com/maki-icons/
+    MGLSymbolStyleLayer *droneLayer = [[MGLSymbolStyleLayer alloc] initWithIdentifier:@"wanderdrone-layer" source:_source];
+    droneLayer.iconImageName = [MGLStyleValue valueWithRawValue:@"rocket-15"];
+    [style addLayer:droneLayer];
+    
+    // Create a timer that calls the `updateUrl` function every 1.5 seconds.
+    _timer = [NSTimer scheduledTimerWithTimeInterval:1.5 target:self selector:@selector(updateURL) userInfo:nil repeats:YES];
```
1ec5: If the developer extends this example to switch styles, it’s possible that `_timer` may already be non-nil by this point. Setting it to a different value will leak the previous timer and this view controller. Similar to `-viewWillDisappear:`, this method also needs to call `-invalidate` on `_timer` before setting it to the new value.

PR #127, opened 2017-11-21
Add tutorial code from mapbox.com/help

captainbarbosa:
Adds the code used for iOS tutorials from [mapbox.com/help/tutorials](mapbox.com/help/tutorials/) into the DocsCode target. This will be used for testing code snippets before being used in documentation on [mapbox.com/help](mapbox.com/help).

### DIFF
--- a/DocsCode/NavigationTutorial/NavigationTutorialViewController.m
+++ b/DocsCode/NavigationTutorial/NavigationTutorialViewController.m
@@ -1,0 +1,155 @@
+// #-code-snippet: navigation full-tutorial-objc
+// #-code-snippet: navigation dependencies-objc
+#import "NavigationTutorialViewController"
+
+@import Mapbox;
+@import MapboxCoreNavigation;
+@import MapboxNavigation;
+@import MapboxDirections;
+// #-end-code-snippet: navigation dependencies-objc
+
+@interface NavigationTutorialViewController () <MGLMapViewDelegate>
+
+// #-code-snippet: navigation mapview-property-objc
+@property (nonatomic) MGLMapView *mapView;
+// #-end-code-snippet: navigation mapview-property-objc
+
+// #-code-snippet: navigation directions-route-objc
+@property (nonatomic) MBRoute *directionsRoute;
+// #-end-code-snippet: navigation directions-route-objc
+
+
+@end
+
+@implementation NavigationTutorialViewController
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    // #-code-snippet: navigation init-map-objc
+    self.mapView = [[MBNavigationMapView alloc] initWithFrame:self.view.bounds styleURL: [MGLStyle streetsStyleURL]];
+    [self.mapView setCenterCoordinate:CLLocationCoordinate2DMake(30.265, -97.741)
+                            zoomLevel:11 animated:NO];
+    [self.view addSubview:self.mapView];
+    // Set the map view's delegate
+    self.mapView.delegate = self;
+    // #-end-code-snippet: navigation init-map-objc
+    
+    // #-code-snippet: navigation user-location-objc
+    // Allow the map view to display the user's location
+    self.mapView.showsUserLocation = YES;
+    // #-end-code-snippet: navigation user-location-objc
+    
+    // #-code-snippet: navigation gesture-recognizer-objc
+    UITapGestureRecognizer *setDestination = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(didLongPress:)];
+    [self.mapView addGestureRecognizer:setDestination];
+    // #-end-code-snippet: navigation gesture-recognizer-objc
+}
+
+// #-code-snippet: navigation allow-callouts-objc
+// Implement the delegate method that allows annotations to show callouts when tapped
+-(BOOL)mapView:(MGLMapView *)mapView annotationCanShowCallout:(id<MGLAnnotation>)annotation {
+    return true;
+}
+// #-end-code-snippet: navigation allow-callouts-objc
+
+// #-code-snippet: navigation long-press-objc
+-(void)didLongPress:(UITapGestureRecognizer *)sender {
+    if (sender.state != UIGestureRecognizerStateEnded) {
+        return;
+    }
+    
+    // Converts point where user did a long press to map coordinates
+    CGPoint point = [sender locationInView:self.mapView];
+    CLLocationCoordinate2D coordinate = [self.mapView convertPoint:point toCoordinateFromView:self.mapView];
+    
+    // Create a basic point annotation and add it to the map
+    MGLPointAnnotation *annotation = [MGLPointAnnotation alloc];
+    annotation.coordinate = coordinate;
+    annotation.title = @"Start navigtation";
+    [self.mapView addAnnotation:annotation];
+    
+    // #-code-snippet: navigation call-calculate-route-objc
+    [self calculateRoutefromOrigin:self.mapView.userLocation.coordinate
+                     toDestination:annotation.coordinate
+                        completion:^(MBRoute * _Nullable route, NSError * _Nullable error) {
+                            if (error != nil) {
+                                printf("Error calculating route");
+                            }
+                        }];
+    // #-end-code-snippet: navigation call-calculate-route-objc
+}
+// #-end-code-snippet: navigation long-press-objc
+
+// #-code-snippet: navigation calculate-route-objc
+-(void)calculateRoutefromOrigin:(CLLocationCoordinate2D)origin
+                 toDestination :(CLLocationCoordinate2D)destination
+                     completion:(void(^)(MBRoute *_Nullable route, NSError *_Nullable error))completion {
+    
+    MBWaypoint *originWaypoint = [[MBWaypoint alloc] initWithCoordinate:origin coordinateAccuracy:-1 name:@"University of Texas at Austin"];
+    
+    MBWaypoint *destinationWaypoint = [[MBWaypoint alloc] initWithCoordinate:destination coordinateAccuracy:-1 name:@"Tacodeli"];
+    
+    MBNavigationRouteOptions *options = [[MBNavigationRouteOptions alloc] initWithWaypoints:@[originWaypoint, destinationWaypoint] profileIdentifier:MBDirectionsProfileIdentifierAutomobileAvoidingTraffic];
+    
+    NSURLSessionDataTask *task = [[MBDirections sharedDirections] calculateDirectionsWithOptions:options completionHandler:^(NSArray<MBWaypoint *> * _Nullable waypoints, NSArray<MBRoute *> * _Nullable routes, NSError * _Nullable error){
+        
+        if (!routes.firstObject) {
+            return;
+        }
+        
+        MBRoute *route = routes.firstObject;
+        self.directionsRoute = route;
+        CLLocationCoordinate2D *routeCoordinates = malloc(route.coordinateCount * sizeof(CLLocationCoordinate2D));
+        [route getCoordinates:routeCoordinates];
+        [self drawRoute:routeCoordinates];
+    }];
+}
+// #-end-code-snippet: navigation calculate-route-objc
+
+// #-code-snippet: navigation draw-route-objc
+-(void)drawRoute:(CLLocationCoordinate2D *)route {
+    if (self.directionsRoute.coordinateCount == 0) {
+        return;
+    }
+    
+    // Convert the route’s coordinates into a polyline.
+    MGLPolylineFeature *polyline = [MGLPolylineFeature polylineWithCoordinates:route count:self.directionsRoute.coordinateCount];
+    
+    if ([self.mapView.style sourceWithIdentifier:@"route-source"]) {
+        // If there's already a route line on the map, reset its shape to the new route
+        MGLShapeSource *source = [self.mapView.style sourceWithIdentifier:@"route-source"];
+        source.shape = polyline;
+    } else {
+        MGLShapeSource *source = [[MGLShapeSource alloc] initWithIdentifier:@"route-source" shape:polyline options:nil];
+        MGLLineStyleLayer *lineStyle = [[MGLLineStyleLayer alloc] initWithIdentifier:@"route-style" source:source];
+        // #-code-snippet: navigation route-style-objc
+        lineStyle.lineColor = [MGLStyleValue valueWithRawValue:[UIColor blueColor]];
+        lineStyle.lineWidth = [MGLStyleValue valueWithRawValue:@"3"];
+        // #-end-code-snippet: navigation route-style-objc
+        [self.mapView.style addSource:source];
+        [self.mapView.style addLayer:lineStyle];
+    }
+    
+}
+// #-end-code-snippet: navigation draw-route-objc
+
+// #-code-snippet: navigation present-navigation-objc
+-(void)presentNavigation:(MBRoute *)route {
+    MBNavigationViewController *viewController = [[MBNavigationViewController alloc] initWithRoute:route
+           directions:[MBDirections sharedDirections]
+                style:nil
+      locationManager:_mapView.locationManager];
+    
+    [self presentViewController:viewController animated:YES completion:nil];
+}
+// #-end-code-snippet: navigation present-navigation-objc
+
+// #-code-snippet: navigation callout-tap-objc
+-(void)mapView:(MGLMapView *)mapView tapOnCalloutForAnnotation:(id<MGLAnnotation>)annotation {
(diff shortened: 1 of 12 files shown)


Podfile
```diff
@@ -1,9 +1,10 @@
-platform :ios, '8.0'
+platform :ios, '9.0'
```
friedbunny: We’ve been trying to keep this at the same minimum deployment target as the SDK, as one more check that we’re not breaking promised backwards compatibility. I’d like to keep it at 8, unless there’s absolutely no way not to.

Podfile
```diff
@@ -1,9 +1,10 @@
...
 use_frameworks!
 
 def shared_pods
   #pod 'Mapbox-iOS-SDK', '~> 3.6.3'
   pod 'Mapbox-iOS-SDK-symbols', :podspec => 'https://raw.githubusercontent.com/mapbox/mapbox-gl-native/ios-v3.6.4/platform/ios/Mapbox-iOS-SDK-symbols.podspec'
+  pod 'MapboxNavigation', '~> 0.10'
```
friedbunny: This is apparently problematic — tests are failing because CocoaPods is trying to install Mapbox.framework twice.

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -0,0 +1,128 @@
+// #-code-snippet: navigation dependencies-swift
+import Mapbox
+import MapboxCoreNavigation
```
friedbunny: Is it necessary to import all four of these? At least `MapboxCoreNavigation` should be able to be implicitly imported.

Podfile
friedbunny: `shared_pods` is installed across all targets, while `MapboxNavigation` is only needed for DocsCode.

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
captainbarbosa: Looks likes it is required from testing it out 😐

captainbarbosa:
It seems like the build is failing because the DocsCode target is using Swift 4. I'll need to wait until https://github.com/mapbox/mapbox-navigation-ios/pull/663 lands and then update the dependency here - once that is done this should pass on BitRise.

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
```diff
@@ -0,0 +1,146 @@
...
+                        }];
+}
+// #-end-code-snippet: navigation long-press-objc
+
+// #-code-snippet: navigation calculate-route-objc
+-(void)calculateRoutefromOrigin:(CLLocationCoordinate2D)origin
+                 toDestination :(CLLocationCoordinate2D)destination
+                     completion:(void(^)(MBRoute *_Nullable route, NSError *_Nullable error))completion {
+    
+    MBWaypoint *originWaypoint = [[MBWaypoint alloc] initWithCoordinate:origin coordinateAccuracy:-1 name:@"Start"];
```
friedbunny: What does `coordinateAccuracy:-1` do? Is `-1` the correct value/type? Is this worth explaining to the reader?

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
```diff
@@ -0,0 +1,146 @@
...
+    
+    MBWaypoint *destinationWaypoint = [[MBWaypoint alloc] initWithCoordinate:destination coordinateAccuracy:-1 name:@"Finish"];
+    
+    // Specify that the route is intented for automobiles avoiding traffic
+    MBNavigationRouteOptions *options = [[MBNavigationRouteOptions alloc] initWithWaypoints:@[originWaypoint, destinationWaypoint] profileIdentifier:MBDirectionsProfileIdentifierAutomobileAvoidingTraffic];
+    
+    // Generate the route object and draw it on the map
+    NSURLSessionDataTask *task = [[MBDirections sharedDirections] calculateDirectionsWithOptions:options completionHandler:^(NSArray<MBWaypoint *> * _Nullable waypoints, NSArray<MBRoute *> * _Nullable routes, NSError * _Nullable error){
```
friedbunny: This is a very long line — can it be broken into something easier to read? The `_Nullable` annotations could be removed, though maybe they’re also useful for a first-timer.

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+        let coordinate = mapView.convert(point, toCoordinateFrom: mapView)
+        
+        // Create a basic point annotation and add it to the map
+        let annotation = MGLPointAnnotation()
+        annotation.coordinate = coordinate
+        annotation.title = "Start navigation"
+        mapView.addAnnotation(annotation)
+        
+        // Calcuate the route from the user's location to the set destination
+        calculateRoute(from: (mapView.userLocation!.coordinate), to: annotation.coordinate) { [unowned self] (route, error) in
```
friedbunny: There’s no use of `self` in this callback, so `[unowned self]` should be unnecessary.

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+                        completion: @escaping (Route?, Error?) -> ()) {
+        
+        let origin = Waypoint(coordinate: origin, name: "Start")
+        let destination = Waypoint(coordinate: destination, name: "Finish")
+        
+        // Specify that the route is intented for automobiles avoiding traffic
+        let options = NavigationRouteOptions(waypoints: [origin, destination], profileIdentifier: .automobileAvoidingTraffic)
+        
+        // Generate the route object and draw it on the map
+        _ = Directions.shared.calculate(options) { (waypoints, routes, error) in
```
friedbunny: `self` _is_ used in this closure — should it be `unowned` or `weak` (or stay `strong`)?

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+            guard let route = routes?.first else { return }
```
friedbunny: You could skip creating this `route` variable and directly assign to `self.directionsRoute` here.

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+        
+        // If there's already a route line on the map, reset its shape to the new route
+        if let source = mapView.style?.source(withIdentifier: "route-source") as? MGLShapeSource {
+            source.shape = polyline
+        } else {
+            let source = MGLShapeSource(identifier: "route-source", features: [polyline], options: nil)
+
+            // Customize the route line color and width
+            let lineStyle = MGLLineStyleLayer(identifier: "route-style", source: source)
+            lineStyle.lineColor = MGLStyleValue(rawValue:  #colorLiteral(red: 0.1897518039, green: 0.3010634184, blue: 0.7994888425, alpha: 1))
```
friedbunny: Nit: looks like there’s an extra space before `#colorLiteral(`.

Podfile
```diff
@@ -10,7 +10,9 @@ target 'Examples' do
 end
 
 target 'DocsCode' do
-  shared_pods
```
friedbunny: We might as well remove `shared_pods`, now that it’s only used for the `Examples` target.

Probably should also switch the `Examples` target to use the non-symbols flavor of our SDK, that way we’re not downloading two different versions for this project.

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
friedbunny: `:squints:` Looks like we missed this one.

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
```diff
@@ -0,0 +1,150 @@
...
+    MBWaypoint *originWaypoint = [[MBWaypoint alloc] initWithCoordinate:origin coordinateAccuracy:-1 name:@"Start"];
+    
+    MBWaypoint *destinationWaypoint = [[MBWaypoint alloc] initWithCoordinate:destination coordinateAccuracy:-1 name:@"Finish"];
+    
+    // Specify that the route is intented for automobiles avoiding traffic
+    MBNavigationRouteOptions *options = [[MBNavigationRouteOptions alloc] initWithWaypoints:@[originWaypoint, destinationWaypoint] profileIdentifier:MBDirectionsProfileIdentifierAutomobileAvoidingTraffic];
+    
+    // Generate the route object and draw it on the map
+    NSURLSessionDataTask *task = [[MBDirections sharedDirections] calculateDirectionsWithOptions:options completionHandler:^(
+            NSArray<MBWaypoint *> *waypoints,
```
friedbunny: These look one indent too far, judging by what follows.

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
```diff
@@ -0,0 +1,150 @@
...
+    [self.mapView addAnnotation:annotation];
+    
+    // Calcuate the route from the user's location to the set destination
+    [self calculateRoutefromOrigin:self.mapView.userLocation.coordinate
+                     toDestination:annotation.coordinate
+                        completion:^(MBRoute * _Nullable route, NSError * _Nullable error) {
+                            if (error != nil) {
+                                printf("Error calculating route");
+                            }
+                        }];
```
friedbunny: This closing `}];` should align with its opening on L71 (`[self calc...`).

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
```diff
@@ -0,0 +1,150 @@
...
+    annotation.coordinate = coordinate;
+    annotation.title = @"Start navigtation";
+    [self.mapView addAnnotation:annotation];
+    
+    // Calcuate the route from the user's location to the set destination
+    [self calculateRoutefromOrigin:self.mapView.userLocation.coordinate
+                     toDestination:annotation.coordinate
+                        completion:^(MBRoute * _Nullable route, NSError * _Nullable error) {
+                            if (error != nil) {
+                                printf("Error calculating route");
```
friedbunny: Use `NSLog()` in ObjC.

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
```diff
@@ -0,0 +1,150 @@
...
+                            }
+                        }];
+}
+// #-end-code-snippet: navigation long-press-objc
+
+// #-code-snippet: navigation calculate-route-objc
+-(void)calculateRoutefromOrigin:(CLLocationCoordinate2D)origin
+                 toDestination :(CLLocationCoordinate2D)destination
```
friedbunny: Nit: extra space before this colon.

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+        // Create a basic point annotation and add it to the map
+        let annotation = MGLPointAnnotation()
+        annotation.coordinate = coordinate
+        annotation.title = "Start navigation"
+        mapView.addAnnotation(annotation)
+        
+        // Calcuate the route from the user's location to the set destination
+        calculateRoute(from: (mapView.userLocation!.coordinate), to: annotation.coordinate) { (route, error) in
+            if error != nil {
+                // Print an error message
```
friedbunny: The ObjC version doesn’t have this comment — it’s probably unnecessary?

Podfile
friedbunny: Thanks. 👍

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
```diff
@@ -0,0 +1,150 @@
...
+    // #-end-code-snippet: navigation init-map-objc
+    
+    // #-code-snippet: navigation user-location-objc
+    // Allow the map view to display the user's location
+    self.mapView.showsUserLocation = YES;
+    // #-end-code-snippet: navigation user-location-objc
+    
+    // #-code-snippet: navigation gesture-recognizer-objc
+    // Add a gesture recognizer to the map view
+    UITapGestureRecognizer *setDestination = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(didLongPress:)];
```
friedbunny: This incorrectly initializes to a `UITapGestureRecognizer ` to `UILongPressGestureRecognizer`.

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
```diff
@@ -0,0 +1,150 @@
...
+    // Coordinate accuracy is the maximum distance away from the waypoint that the route may still be considered viable, measured in meters. Negative values indicate that a indefinite number of meters away from the route and still be considered viable.
+    MBWaypoint *originWaypoint = [[MBWaypoint alloc] initWithCoordinate:origin coordinateAccuracy:-1 name:@"Start"];
+    
+    MBWaypoint *destinationWaypoint = [[MBWaypoint alloc] initWithCoordinate:destination coordinateAccuracy:-1 name:@"Finish"];
+    
+    // Specify that the route is intented for automobiles avoiding traffic
+    MBNavigationRouteOptions *options = [[MBNavigationRouteOptions alloc] initWithWaypoints:@[originWaypoint, destinationWaypoint] profileIdentifier:MBDirectionsProfileIdentifierAutomobileAvoidingTraffic];
+    
+    // Generate the route object and draw it on the map
+    NSURLSessionDataTask *task = [[MBDirections sharedDirections] calculateDirectionsWithOptions:options completionHandler:^(
```
friedbunny: This line triggers the warning, `Unused variable “task”`.

Until https://github.com/mapbox/MapboxDirections.swift/pull/209 lands, the way to avoid this is to explicitly discard the result, like so:

```objc
(void)[[[MBDirections sharedDirections] calc...
```

DocsCode/NavigationTutorial/NavigationTutorialViewController.m
friedbunny: Also, since we have the `error` object, let’s print that, too. ([Like this.](https://github.com/mapbox/MapboxDirections.swift/blob/11313a6ce9909117992a6655f79a4e739ed93f30/Directions%20Example/ViewController.m#L41).

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+                print("Error calculating route")
```
friedbunny: [Same.](https://github.com/mapbox/ios-sdk-examples/pull/127/files#r153367342)

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -58,7 +58,7 @@ class ViewController: UIViewController, MGLMapViewDelegate {
         // Calcuate the route from the user's location to the set destination
         calculateRoute(from: (mapView.userLocation!.coordinate), to: annotation.coordinate) { (route, error) in
             if error != nil {
-                print("Error calculating route")
+                NSLog("Error calculating route: \(String(describing: error))")
```
friedbunny: So `print()` _is_ the recommended way to put text to console in Swift. 😁 It’s possible to use `NSLog()` in Swift (and there are [behavioral differences](https://stackoverflow.com/a/25951564/2094275) that aren’t relevant here), but `print()` is preferred.

The issue with the ObjC code was in using `printf()`, which is a C function.

DocsCode/NavigationTutorial/NavigationTutorialViewController.swift
```diff
@@ -0,0 +1,121 @@
...
+        
+        // Coordinate accuracy is the maximum distance away from the waypoint that the route may still be considered viable, measured in meters. Negative values indicate that a indefinite number of meters away from the route and still be considered viable.
+        let origin = Waypoint(coordinate: origin, coordinateAccuracy: -1, name: "Start")
+        let destination = Waypoint(coordinate: destination, coordinateAccuracy: -1, name: "Finish")
+        
+        // Specify that the route is intented for automobiles avoiding traffic
+        let options = NavigationRouteOptions(waypoints: [origin, destination], profileIdentifier: .automobileAvoidingTraffic)
+        
+        // Generate the route object and draw it on the map
+        Directions.shared.calculate(options) { (waypoints, routes, error) in
```
friedbunny: We need `_ =` to avoid a warning about an unused return value. We can remove it when https://github.com/mapbox/MapboxDirections.swift/pull/209 makes it into a release.

PR #135, opened 2017-12-11
Rename examples

captainbarbosa:
Renames a handful of examples to better describe their purpose.


captainbarbosa:
@friedbunny @1ec5 - there are a handful of examples I couldn't think of a good name for. Let me know if ya'll have ideas on:

- `DrawingAGeoJSONLineExample`
- `RuntimeAddLineExample`
- `DDSCircleLayerExample`

Examples/Swift/ShowHideLayerExample.swift
```diff
@@ -74,7 +74,7 @@ class RuntimeToggleLayerExample_Swift: UIViewController, MGLMapViewDelegate {
         button.sizeToFit()
         button.center.x = self.view.center.x
         button.frame = CGRect(origin: CGPoint(x: button.frame.origin.x, y: self.view.frame.size.height - button.frame.size.height - 5), size: button.frame.size)
-        button.addTarget(self, action: #selector(RuntimeToggleLayerExample_Swift.toggleLayer(sender:)), for: .touchUpInside)
+        button.addTarget(self, action: #selector(ShowHideLayerExample_Swift.toggleLayer(sender:)), for: .touchUpInside)
```
friedbunny: Let’s try to remove the class name from this selector — it’s an implementation detail that breaks direct copy-pasting.

Examples/ObjectiveC/StudioClassicStyleExample.m
```diff
@@ -1,9 +1,9 @@
-#import "CustomRasterStyleExample.h"
+#import "StudioClassicStyleExample.h"
```
friedbunny: This isn’t entirely accurate — it can be used with any XYZ rasters, not just ones created with Studio Classic or hosted with Mapbox. The note [on the examples page](https://www.mapbox.com/ios-sdk/examples/raster-styles/) makes that clear.

1ec5: If I’m not mistaken, the example’s title would mention Studio Classic to avoid the (over)use of the word “custom”. Perhaps “raster style” is good enough?

captainbarbosa: Yeah, this was changed to reflect the upcoming change we're making on the live examples page, since it will soon be `Apply a style designed in Mapbox Studio Classic` instead of `Custom raster style` - I would rather have the titles match similarly between the examples page and the ones that live in here.

captainbarbosa:
I've think I'm done with renaming all the examples I wanted to. Let me know if there are other that could use a new name :)

friedbunny:
> This would be useful but we'd also have to keep them up to date between the two if they get moved around - I'm not sure if its worth it.

Let’s do it.

friedbunny:
@captainbarbosa I left the new folders on disk intact — I suppose it’ll be worth the effort to maintain if everything is structured in the same way.

captainbarbosa:
Thanks for your help with the reorganizing @friedbunny 😄  I also went ahead and fixed the merge conflicts. It would be great if you or @1ec5 could do a final review.

PR #137, opened 2017-12-13
Add Navigation Hello World example

bsudekum:
This adds a simple hello world example in Swift. There were a couple of semi-invasive changes:

- Changing the min deployment target from 8 -> 9.
- added UIBackgroundModes to the plist

Per a conversation yesterday, while we work on putting together a site similar to `mapbox.com/ios-sdk/`, we thought this would be a good home.

I still need to create an objc example along with this, but I thought I'd get the conversation going on this change first.

/cc @friedbunny @captainbarbosa @ericrwolfe @colleenmcginnis


friedbunny:
Let’s create a new target within the existing project specifically for navigation, given that it has specialized needs that conflict with the current Examples target.

PR #139, opened 2017-12-18
Remove example subgroup folders

captainbarbosa:
The decision to place examples in subgroup folders by topic causes downstream issues for internal documentation systems, so this PR removed them from their subfolders and instead holds all examples in their original `Swift` and `Objective-C` directories.


friedbunny:
Can we keep the folders in the project but keep it flat on disk? With so many examples it’s become difficult to navigate a long list and having them in consistent groups makes that easier.

captainbarbosa:
Yeah, it looks like that is possible with Xcode 9 with some poking and prodding through the new interface. PR for that is here: https://github.com/mapbox/ios-sdk-examples/pull/140

PR #152, opened 2018-02-28
DraggableAnnotationView: add haptic feedback

friedbunny:
Adds haptic feedback to the drag-and-drop annotations example.

Caveat: ObjC lets us declare properties with potentially unavailable classes and then check availability before initialization, but Swift does not. So, the Swift version can’t follow Apple’s suggested best practice of reusing the feedback generator (but the ObjC version can and does).

More about feedback generators can be found [at Apple’s docs](https://developer.apple.com/documentation/uikit/uifeedbackgenerator).

/cc @captainbarbosa @jmkiley @1ec5

### DIFF
--- a/Examples/Swift/DraggableAnnotationViewExample.swift
+++ b/Examples/Swift/DraggableAnnotationViewExample.swift
@@ -19,7 +19,7 @@ class DraggableAnnotationViewExample_Swift: UIViewController, MGLMapViewDelegate
         let coordinates = [
             CLLocationCoordinate2D(latitude: 0, longitude: -70),
             CLLocationCoordinate2D(latitude: 0, longitude: -35),
-            CLLocationCoordinate2D(latitude: 0,  longitude: 0),
+            CLLocationCoordinate2D(latitude: 0, longitude: 0),
             CLLocationCoordinate2D(latitude: 0, longitude: 35),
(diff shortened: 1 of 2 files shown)


Examples/Swift/DraggableAnnotationViewExample.swift
```diff
@@ -115,6 +115,12 @@ class DraggableAnnotationView: MGLAnnotationView {
             self.layer.opacity = 0.8
             self.transform = CGAffineTransform.identity.scaledBy(x: 1.5, y: 1.5)
         }, completion: nil)
+
+        // Initialize haptic feedback generator and give the user a light thud.
+        if #available(iOS 10.0, *) {
+            let hapticFeedback = UIImpactFeedbackGenerator(style: .light)
```
jmkiley: Should we add a note about not being able to store the potentially unavailable object as a property?

friedbunny: I dunno — it's a system limitation that is annoying but isn’t ultimately a huge deal, so we don’t strictly have to mention it. It verges on being a bit too meta, I’d say.

If anyone feels strongly about it and can think of a good way to say, “Typically you’d reuse the feedback generator but we can’t because Swift and backwards compatibility” — please feel free. 🤔

PR #154, opened 2018-03-02
[WIP] Symbol layer with multiple images

jmkiley:
This is very much a work in progress.


jmkiley:
Current example:
simulator screen shot - iphone 7 - 2018-03-08 at 13 36 02

jmkiley:
This currently uses 4.0.0-beta.2, and will be maintained with the rest of the v4.0.0-expressions branch. [Icons](https://www.nps.gov/hfc/carto/map-symbols.cfm) and [data](https://irma.nps.gov/DataStore/Reference/Profile/2225064) from the NPS.

Examples/Assets.xcassets/nps-picnic-area.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "32px-Pictograms-nps-picnic_area.svg.png",
```
friedbunny: Since we’re evidently sourcing these from SVG, let’s use PDF here instead of a single PNG — that way we only have to provide one file to cover all scales.

friedbunny: It may also make sense to put these in their own `xcassets` catalog, so that we can have developers download that directly.

Examples/ObjectiveC/MultipleImagesExample.m
```diff
@@ -0,0 +1,52 @@
...
+    [self.view addSubview:mapView];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    // Add icons from the U.S. National Parks Service to the map's style.
+    [style setImage:[UIImage imageNamed:@"nps-restrooms"] forName:@"restrooms"];
+    [style setImage:[UIImage imageNamed:@"nps-trailhead"] forName:@"trailhead"];
+    [style setImage:[UIImage imageNamed:@"nps-picnic-area"] forName:@"picnic-area"];
+    
+    NSURL *url = [[NSURL alloc] initWithString:@"mapbox://examples.ciuz0vpc"];
```
friedbunny: We might add some explanation of what this source is and how someone could create their own.

Minor point, but the `url` variable name could stand to be a little more descriptive, too.

Examples/ObjectiveC/MultipleImagesExample.m
```diff
@@ -0,0 +1,52 @@
...
+    MGLVectorSource *source = [[MGLVectorSource alloc] initWithIdentifier:@"yosemite-pois" configurationURL:url];
+    [style addSource:source];
+    
+    MGLSymbolStyleLayer *layer = [[MGLSymbolStyleLayer alloc] initWithIdentifier:@"yosemite-pois" source:source];
+    
+    // The source name from the source's TileJSON metadata: mapbox.com/api-documentation/#retrieve-tilejson-metadata
```
friedbunny: Is this link clickable?

jmkiley:
Renamed the icon files and moved them to a separate assets collection. I think I'm going to keep the `PNG` files for now. cc @friedbunny

friedbunny:
> I think I'm going to keep the `PNG` files for now.

Then we’ll need to add PNGs for the other scales, otherwise `@2x` and `@3x` will be awkwardly upscaled.

jmkiley:
Thanks! Added them.

jmkiley:
Will convert to v4.0.0-beta.3

Issue #166, opened 2018-04-19
Example list should be better organized and more friendly


friedbunny:
The list of examples in the initial table view should:

- Use the prose name of the example, per the website.
- Be sectioned by category, per the organization on the website.
- Have one entry per example, perhaps containing separate buttons for Obj-C and Swift.

<img src=https://user-images.githubusercontent.com/1198851/39009560-fbbad7f6-43d9-11e8-84d8-aa87aae9948e.jpeg width=300>

/cc @jmkiley @captainbarbosa @lilykaiser

lloydsheng:
@friedbunny  That's what i working at! I was planing cut up a ticket and saw this issue. 😀
Last week i've prepared a demo app for Wework Shanghai Job fair which based on `ios-sdk-examples`. But I think it difficult to find the example I want through looking up controller name list. So i organized examples with category names and add some brief descriptions, here are the screenshots:

### iPad Landscape:
![img_0018](https://user-images.githubusercontent.com/503040/39037222-6ca1188a-44b2-11e8-994b-45ed968fd99f.jpg)

### iPad Portrait:
![img_0019 3](https://user-images.githubusercontent.com/503040/39037432-125b67bc-44b3-11e8-9b4a-1de9b92ab1ff.PNG)

## iPhone:
![img_0002](https://user-images.githubusercontent.com/503040/39037509-4c3eae08-44b3-11e8-88ee-8dee0a522e36.PNG)

 I'm going to:

- [ ] Category name, title, description  localization(English, Chinese)

- [ ] Create a pull request for those changes.

Look forward to your thoughts.

lloydsheng:
@lilykaiser @chriswu42   Video demos:

https://www.dropbox.com/s/v1ghsi1wjp1swxx/Examples-iPad.mp4?dl=0
https://www.dropbox.com/s/q9ktoyaf16ii6ex/Examples-iPhone.mp4?dl=0

lilykaiser:
Looks great!

jmkiley:
Closed via #172

PR #172, opened 2018-04-26
[WIP]Example list should be better organized and more friendly

lloydsheng:
Per https://github.com/mapbox/ios-sdk-examples/issues/166
- [x]  Improve iPad user experience.  `UISplitViewController`
- [x]  Reorganize examples with categories 
- [ ]  China Localization

### DIFF
--- a/Examples/Examples.m
+++ b/Examples/Examples.m
@@ -33,7 +33,7 @@
             ]
         },
         @{
-            @"title": @"Markers and callouts",
+            @"title": @"Annotations",
             @"examples":@[
                     @{@"className": MBXExampleCustomAnnotationModel, @"title": @"Annotation models"},
                     @{@"className": MBXExampleAnnotationViewsAndImagesExample, @"title": @"Add annotation views and images"},
--- a/Examples/Examples.m
+++ b/Examples/Examples.m
@@ -10,63 +10,127 @@
 
 @implementation Examples
 
-+ (NSArray *)list {
-    
++ (NSArray *)groups {
     NSArray *initialList = [[NSMutableArray alloc] initWithArray:@[
-        MBXExampleAnimatedLine,
-        MBXExampleAnnotationView,
-        MBXExampleAnnotationViewsAndImagesExample,
-        MBXExampleBlockingGesturesDelegate,
-        MBXExampleBuildingLight,
-        MBXExampleBuildings,
-        MBXExampleCalloutDelegateUsage,
-        MBXExampleCameraAnimation,
-        MBXExampleCameraFlyTo,
-        MBXExampleClustering,
-        MBXExampleCustomAnnotationModel,
-        MBXExampleCustomCalloutView,
-        MBXExampleDDSCircleLayer,
-        MBXExampleDefaultAnnotation,
-        MBXExampleDefaultStyles,
-        MBXExampleDraggableAnnotationView,
-        MBXExampleFeatureSelection,
-        MBXExampleImageAnnotation,
-        MBXExampleImageSource,
-        MBXExampleLineAnnotationGeoJSON,
-        MBXExampleLineStyleLayer,
-        MBXExampleLiveData,
-        MBXExampleMultipleImages,
-        MBXExampleMultipleShapes,
-        MBXExampleOfflinePack,
-        MBXExamplePointConversion,
-        MBXExamplePointHotspot,
-        MBXExamplePolygonAnnotation,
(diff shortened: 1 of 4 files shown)


Examples/Examples.m
```diff
@@ -10,63 +10,127 @@
...
+                    @{@"className": MBXExampleAnnotationView, @"title": @"Annotation views"},
+                    @{@"className": MBXExampleCalloutDelegateUsage, @"title": @"Callout delegate usage"},
+                    @{@"className": MBXExampleCustomCalloutView, @"title": @"Display custom views as callouts"},
+                    @{@"className": MBXExampleImageAnnotation, @"title": @"Mark a place on the map with an image"},
+                    @{@"className": MBXExampleWebAPIData, @"title": @"Dynamically style interactive points"},
+                    @{@"className": MBXExampleUserLocationAnnotation, @"title": @"Customize the user location annotation"},
+                    ]
+        },
+        @{
+            @"title": @"Markers and callouts",
```
frederoni: nit: duplicate of previous title, s/Markers and callouts/Annotations

friedbunny:
@lloydsheng I missed your original post in the recent release bustle, but thanks for working on this.

As for the direction this PR should take, I’m not sure — any work should be done in coordination/consultation with @jmkiley and @captainbarbosa, who are the primary stewards of this repo and likely have thoughts about its future direction.

For instance:
- How to bring in title/description strings from `ios-sdk`?
- Is localizing this project something we want to maintain in the long run?

lloydsheng:
> How to bring in title/description strings from ios-SDK?

We cloud sync titles/descriptions with https://www.mapbox.com/ios-sdk/examples/

> Is localizing this project something we want to maintain in the long run?

As far as i know, currently we should just maintain Chinese localization and China team can take over this work.

@friedbunny  Thanks for your suggestions, I'm not familiar with the workflow of `ios-sdk-examples`. @jmkiley @captainbarbosa please tell me your thoughts about this, I will cut up a ticket if we need further discussion.

captainbarbosa:
@chriswu42 @1ec5 suggested maintaining a different branch on this repo for maintaining Chinese localization - a separate repository might be difficult to maintain. 

@lloydsheng this is a nice improvement! One issue I could see us running into is that not every example in this project is listed on our site. For example, the [`CameraFlyToExample.swift`](https://github.com/mapbox/ios-sdk-examples/blob/b330e07afa5b177f83ccb5aa2e05854a2c3e0589/Examples/Swift/CameraFlyToExample.swift) (and its Objective-C counterpart) isn't on our [examples page](https://www.mapbox.com/ios-sdk/examples/) but its in this project. When a new example is created here, we have to manually make a new page for it and assign it a title after the example for it has been created here so it can live on https://www.mapbox.com/ios-sdk/examples/.

lloydsheng:
> This is a nice improvement! One issue I could see us running into is that not every example in this project is listed on our site.

@captainbarbosa  yes, here is iOS examples workflow i understood so far.
1. Implemented an example in `ios-SDK-examples`
2. Determine if we should write an awesome document about it (Which aspects we consider?)
3. If yes, the documentation should be in https://www.mapbox.com/ios-sdk/examples/

If my understanding is right, Number of examples inside `ios-sdk-examples` would more than documents.
I insist that we could think about the first two changes and merge them to master branch. because it's well organized and user-friendly in iPad iPhone. 🤓
About the China localization, Maintain it in an exclusive branch (cn-examples) is ok to me.

/cc @chriswu42

chriswu42:
@lloydsheng are you planning to create this china branch?

@captainbarbosa anything we should be mindful of in maintaining this branch?

captainbarbosa:
I'm 👍 with the proposed changes. 

In terms of keeping a `cn-examples` branch up-to-date, we regularly update the Podfile when new versions are released - this is part of the Maps SDK release checklist, but you'll have to pull in those changes when they merge to keep it up to date.

@lloydsheng in response to your question:

>Determine if we should prepare an awesome documentation presenting it (What are aspects we consider?)

Generally, when considering adding an example on the live page, proposed examples should:
- Should demonstrate one particular feature.
- Have minimal UIKit code so that buttons, switches, and sliders and the like don't distract from the example's goal.
-  Is not similar to other examples already illustrated on the page.

lloydsheng:
>  are you planning to create this china branch?
@chriswu42  Yes, I've created  an `cn-examples` branch. Currently. It will support China localization and the work is in progress.

PR #174, opened 2018-05-02
Switched to using bounds rather than frame for cornerRadius (as frame changes when rotated).

julianrex:
Partly addresses https://github.com/mapbox/mapbox-gl-native/issues/11756 (also added `rotatesToMatchCamera = YES` to test).

See also https://github.com/mapbox/mapbox-gl-native/pull/11817 for the associated change in mapbox-gl-native.

### DIFF
--- a/Examples/ObjectiveC/AnnotationViewExample.m
+++ b/Examples/ObjectiveC/AnnotationViewExample.m
@@ -12,10 +12,6 @@ NSString *const MBXExampleAnnotationView = @"AnnotationViewExample";
 - (void)layoutSubviews {
     [super layoutSubviews];
 
-    // Force the annotation view to maintain a constant size when the map is tilted.
-    self.scalesWithViewingDistance = NO;
-    self.rotatesToMatchCamera = YES;
-
     // Use CALayer’s corner radius to turn this view into a circle.
     self.layer.cornerRadius = self.bounds.size.width / 2;
     self.layer.borderWidth = 2;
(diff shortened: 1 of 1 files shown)


Examples/ObjectiveC/AnnotationViewExample.m
```diff
@@ -13,10 +13,11 @@ - (void)layoutSubviews {
     [super layoutSubviews];
 
     // Force the annotation view to maintain a constant size when the map is tilted.
-    self.scalesWithViewingDistance = false;
+    self.scalesWithViewingDistance = NO;
+    self.rotatesToMatchCamera = YES;
```
friedbunny: Since this is a relatively basic example, let’s remove this test code before merging (otherwise we may encourage cargo-culting, as this setting should have no effect for a round annotation).

Incidentally, now that `scalesWithViewingDistance` is off by default, we should remove that, too.

julianrex: How about making one of those round annotations square? 😉

friedbunny: When thinking about changing examples, I like to consider:

- Does this help illustrate the underlying concept?
- Is this complexity necessary?
- Does this visual change justify the effort of re-recording the video for the site?

These examples are definitely useful as tests, but are primarily meant as learning tools for developers unfamiliar with our SDK — for test code that you want to keep around (but not turn into a formal test), iosapp is our dumping ground for that. 😉

jmkiley: Capturing from chat that we're going to look at moving the tests and notifications to a separate view controller. Testing examples in general would be useful for catching smaller issues that arise.

julianrex: > Does this visual change justify the effort of re-recording the video for the site?

Ah ha! I hadn't consider videos. Very good point.

PR #175, opened 2018-05-03
Initial XCUITest handling

julianrex:
In the following PR:
- I've tried to leave the examples untouched; any test related code is in an associated category.
- Communication to the host test is via notifications (or existence of UI elements, but non of these tests do that). To handle this, this code inserts a forwarding delegate to manage posting notifications.
- One issue is that this delegate is inserted in `viewWillAppear` implemented in the category - this is fine right now, as NONE of the examples currently implement it. But ideally there would be a better way to manage this. Thoughts?
- This just uses XCUITests, without any 3rd party dependencies. Ideally, we'd keep it like this.
- I rejigged the `testEveryExample` test to use `XCTActivity` so that the test report shows a clean list of tests. It means the implementation is a little convoluted, but results in a cleaner log.

Thoughts?

/cc @akitchen


julianrex:
BTW I've only added 3 tests (picked somewhat at random) - ideally we'll add more as time goes on (perhaps as we're tackling bugs).

julianrex:
Updated for recent UI changes and Podfile.lock. @jmkiley can you take a look at this one

julianrex:
Before I merge, @akitchen would you mind taking a look to see if there's a cleaner way to do this - thanks!

akitchen:
That's an interesting way of hooking in. Looks OK to me, so long as the mapVIew's delegate isn't expected to change... I'm not sure you need the optimization in `-respondsToSelector:` but it's probably fine.

julianrex:
Interestingly, without that optimization test times were severely affected! Much more than I had been expecting. 

I'll go ahead and merge, but let's continue to iterate on this. @jmkiley and I will discuss.

PR #186, opened 2018-06-07
CN UI localization (category name/sample name)

lloydsheng:
I think UI localization is a common feature, this change should merge to `master` branch. 
We also have a `cn-examples` branch which is a better place for putting China specific examples, such as China Plugin examples.

/cc @chriswu42


lloydsheng:
@captainbarbosa @jmkiley May you take some time review this change  🙇

jmkiley:
:wave:  It may be better for this PR to live on the `cn-examples` branch. @captainbarbosa and I wouldn’t be able to add accurate Chinese titles to the new examples, and we are trying to keep the examples app fairly bare bones.

lloydsheng:
@jmkiley Thanks for the suggestion, that makes sense to me, will move to `cn-examples` branch.

Issue #177, opened 2018-05-11
Show how to use an MGLCircleStyleLayer with MGLTransition


captainbarbosa:
An example of how to use MGLCircleStyleLayer with MGLTransitions would be a solid way to illustrate how to gradually change how a circle layer is styled when selected.

riastrad:
Not that this example is on hold until the updates discussed in mapbox/mapbox-gl-js#3170 are implemented.

PR #187, opened 2018-06-12
[WIP DO NOT MERGE] Use MGLTransition to animate MGLCircleStyleLayer 

riastrad:
Adding examples to demonstrate possible use, per #177.

cc: @captainbarbosa & @jmkiley


riastrad:
Sadly, after digging into this a bit more, it appears that this is not something that can be implemented using MGLTransition at this time. Transitions do not currently support source style expressions (see: mapbox/mapbox-gl-native#10666 & mapbox/mapbox-gl-js#3170).

As a result, I'm shelving this example until a cleaner implementation is possible.

PR #193, opened 2018-06-28
Update podspecs to 4.1.1.

fabian-guerra:
Bumps podspec to 4.1.1

PR #195, opened 2018-07-02
Update the name/description for the default callout example

jmkiley:
Changed the name of the `Callout Delegate Usage Example` to `Default Callout Example`. The example doesn't seem to use any [MGLCalloutViewDelegate](https://www.mapbox.com/ios-sdk/api/4.1.1/Protocols/MGLCalloutViewDelegate.html) methods.


Podfile.lock
```diff
@@ -1,12 +1,12 @@
 PODS:
-  - Mapbox-iOS-SDK (4.1.0)
+  - Mapbox-iOS-SDK (4.1.1)
```
captainbarbosa: Should we be updating to 4.1.1 in this PR?

jmkiley: Thanks! Looks like there is an open PR to address that (#193)

jmkiley:
Removed the `Podfile.lock` from this PR.

PR #196, opened 2018-07-03
Add enlarge annotation example

### LINKED ISSUE #189: Add enlarge annotation example
Both Google and Apple have made the "select an annotation, make it larger" type animation pretty popular in years past:

Let's add an example that does just this. Nothing too complicated, but should 

1. enlarge the selected annotation
1. shrink any enlarged annotation when deselecting

Maybe we can even talk about adding this into the sdk proper?

/cc @jmkiley @1ec5

jmkiley:
I updated the annotation view example to scale the selected annotation. I ran into an issue where the positioning of the callout view is not updated based on the new size of the selected annotation. 

My current workaround for this is to call `mapView.setNeedsLayout` in `mapView:didSelectAnnotation`.  It is a little jumpy, though. 

Fixes #189 

cc @bsudekum

### DIFF
--- a/Examples/ObjectiveC/AnnotationViewExample.m
+++ b/Examples/ObjectiveC/AnnotationViewExample.m
@@ -108,7 +108,7 @@ NSString *const MBXExampleAnnotationView = @"AnnotationViewExample";
 
 // Adjust the positioning of the callout view once the annotation is selected.
 - (void)mapView:(MGLMapView *)mapView didSelectAnnotation:(id<MGLAnnotation>)annotation {
-    [mapView layoutSubviews];
+    [mapView setNeedsLayout];
 }
 
 @end
--- a/Examples/ObjectiveC/AnnotationViewExample.m
+++ b/Examples/ObjectiveC/AnnotationViewExample.m
@@ -21,11 +21,14 @@ NSString *const MBXExampleAnnotationView = @"AnnotationViewExample";
 - (void)setSelected:(BOOL)selected animated:(BOOL)animated {
     [super setSelected:selected animated:animated];
 
-    // Animate the border width in/out, creating an iris effect.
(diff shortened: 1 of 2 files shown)


Examples/ObjectiveC/AnnotationViewExample.m
```diff
@@ -21,11 +21,14 @@ - (void)layoutSubviews {
...
     [super setSelected:selected animated:animated];
 
-    // Animate the border width in/out, creating an iris effect.
-    CABasicAnimation *animation = [CABasicAnimation animationWithKeyPath:@"borderWidth"];
-    animation.duration = 0.1;
-    self.layer.borderWidth = selected ? self.bounds.size.width / 4 : 2;
-    [self.layer addAnimation:animation forKey:@"borderWidth"];
+    // Increase the size of the annotation if it is selected, and restore annotation to its original size if it is not selected.
+    CABasicAnimation *animation = [CABasicAnimation animationWithKeyPath:@"bounds.size"];
+    [self.layer addAnimation:animation forKey:@"bounds.size"];
```
julianrex: In general we should prefer the higher level animation constructs where possible. Let's look at whether something like
```
[UIView animateWithDuration:kSelectionAnimationDuration animations:^{ 
    self.transform = CGAffineTransformMakeScale(2.0, 2.0);
}];
```
would be more appropriate. (We still need to support iOS 9, otherwise I think the new `UIViewPropertyAnimator` would be preferable.)

In addition, in the above you add an animation for `bounds.size` - but when you scale by modifying the transform `bounds` doesn't change size (`frame` is a computed property and would).

julianrex: We probably want to check the `animated` method parameter too - to decide whether the scale should be animated.

friedbunny:
Is this a thing we should finish?

julianrex:
Yep, I think so.

Issue #185, opened 2018-05-29
Clustering with images


captainbarbosa:
An example of how to use clustering with [`MGLSymbolStyleLayer.iconImageName`](https://www.mapbox.com/ios-sdk/api/4.0.1/Classes/MGLSymbolStyleLayer.html#/c:objc(cs)MGLSymbolStyleLayer(py)iconImageName) instead of `MGLCircleStyleLayer` would be helpful as I've seen a couple of requests for it before.

I have roughly done this in the past for a couple of internal examples before, but for demonstration purposes the images should be be remade and the Objective-C example is still required.

```swift
let numbersLayer = MGLSymbolStyleLayer(identifier: "clusteredPortsNumbers", source: source)
numbersLayer.textColor = NSExpression(forConstantValue: UIColor.white)
numbersLayer.textFontSize = NSExpression(forConstantValue: NSNumber(value: Double(icon.size.width) / 2))
numbersLayer.iconAllowsOverlap = NSExpression(forConstantValue: true)
        
// CLUSTER STYLING
        
style.setImage(UIImage(named: "circle")!, forName: "circle")
style.setImage(UIImage(named: "rectangle")!, forName: "rectangle")
style.setImage(UIImage(named: "star")!, forName: "star")
style.setImage(UIImage(named: "polygon")!, forName: "polygon")
        
let stops = [
        20:  NSExpression(forConstantValue: "circle"),
        50:  NSExpression(forConstantValue: "rectangle"),
        100: NSExpression(forConstantValue: "star"),
        200: NSExpression(forConstantValue: "polygon")
    ]
        
let defaultStar = NSExpression(forConstantValue: "star")
numbersLayer.iconImageName = NSExpression(format: "mgl_step:from:stops:(point_count, %@, %@)", defaultStar, stops)
numbersLayer.text = NSExpression(format: "CAST(point_count, 'NSString')")
```

riastrad:
@jmkiley @captainbarbosa do you think that including the `singleTap` functionality that's present in the current `MGLCircleStyleLayer` [clustering example](https://github.com/mapbox/ios-sdk-examples/blob/master/Examples/Swift/ClusteringExample.swift#L20-L25) is necessary for this one? 

[...]

PR #199, opened 2018-07-30
[WIP] Cluster with Images Example (#185)

riastrad:
As @captainbarbosa pointed out, there is a need for an example that demonstrates how to manage clusters using MGLSymbolStyleLayer.iconImageName. This PR satisfies that requirement.


riastrad:
img_0159

Not thrilled with how text-anchor is currently rendering on the polygon & star images. I suspect there's an easy solve for this, but need to do some digging.

riastrad:
@jmkiley @captainbarbosa, this one's ready to review whenever either of you have time!

Examples/ObjectiveC/ClusteringWithImagesExample.m
```diff
@@ -0,0 +1,69 @@
+//
+//  ClusteringWithImagesExample.m
+//  Examples
+//
+//  Created by Joshua Erb on 7/30/18.
+//  Copyright © 2018 Mapbox. All rights reserved.
+//
+
```
captainbarbosa: I think we generally remove these comments from the top of example files.

Examples/ObjectiveC/ClusteringWithImagesExample.m
```diff
@@ -0,0 +1,69 @@
...
+    [style setImage:[UIImage imageNamed:@"rectangle"] forName:@"rectangle"];
+    [style setImage:[UIImage imageNamed:@"cloud"] forName:@"cloud"];
+    [style setImage:[UIImage imageNamed:@"oval"] forName:@"oval"];
+    
+    NSDictionary *stops = @{ @10: [NSExpression expressionForConstantValue:@"circle"],
+                             @25: [NSExpression expressionForConstantValue:@"rectangle"],
+                             @75: [NSExpression expressionForConstantValue:@"cloud"],
+                             @150: [NSExpression expressionForConstantValue:@"oval"] };
+    
+    NSExpression *defaultCircle = [NSExpression expressionForConstantValue:@"circle"];
```
captainbarbosa: I recommend using a different image for the default circle so it doesn't get mixed up with the circle defined in the stops dictionary.

Examples/ObjectiveC/ClusteringWithImagesExample.m
```diff
@@ -0,0 +1,69 @@
...
+    self.mapView.delegate = self;
+    [self.view addSubview:self.mapView];
+    
+    self.icon = [UIImage imageNamed:@"circle"];
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    NSURL *url = [NSURL fileURLWithPath:[[NSBundle mainBundle] pathForResource:@"ports" ofType:@"geojson"]];
+    
+    // Retrieve data and set as style layer source
```
captainbarbosa: Can add more comments on sources/style layers here and in the lines below? It might be useful to mention the source data is used to load data into the map but style layers actually configure visual styling to the data. It might be also good to clarify that three different style layers are required for this whole thing: a layer to display unclustered points, a layer to display clustered images, and a layer to display labels for the clusters.

Side note, can you include the unclustered points too, like we do in the original clustering example?

```objc
    // Show unclustered features as icons. The `cluster` attribute is built into clustering-enabled source features.
    MGLSymbolStyleLayer *ports = [[MGLSymbolStyleLayer alloc] initWithIdentifier:@"ports" source:source];
    ports.iconImageName = [NSExpression expressionForConstantValue:@"icon"];
    ports.iconColor = [NSExpression expressionForConstantValue:[[UIColor darkGrayColor] colorWithAlphaComponent:0.9]];
    ports.predicate = [NSPredicate predicateWithFormat:@"cluster != YES"];
    [style addLayer:ports];
```

Examples/ObjectiveC/ClusteringWithImagesExample.m
```diff
@@ -0,0 +1,69 @@
...
+    [super viewDidLoad];
+    
+    // Initialize the Map
+    self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds
+                                            styleURL:[MGLStyle lightStyleURL]];
+    self.mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    self.mapView.delegate = self;
+    [self.view addSubview:self.mapView];
+    
+    self.icon = [UIImage imageNamed:@"circle"];
```
captainbarbosa: Can you move this line to the `didFinishLoadingStyle` delegate method? Since it isn't it used until then, I'm not sure if setting it all up in the `viewDidLoad` is necessary.

Another option would be to hard-code these values, then you wouldn't need to include the `icon` property on the ViewController.

Podfile
```diff
@@ -2,7 +2,7 @@ platform :ios, '9.0'
 use_frameworks!
 
 def shared_pods
-    pod 'Mapbox-iOS-SDK', '~> 4.1.1'
+    pod 'Mapbox-iOS-SDK', '~> 4.2.0'
```
captainbarbosa: Can you drop back to 4.1.1? While we do want to update this, I think it should be done in separate PR.

Examples/ObjectiveC/ClusteringWithImagesExample.m
```diff
@@ -0,0 +1,69 @@
...
+                              MGLShapeSourceOptionClusterRadius: @(self.icon.size.width)
+    }];
+    [style addSource:source];
+    
+    MGLSymbolStyleLayer *numbersLayer = [[MGLSymbolStyleLayer alloc] initWithIdentifier:@"clusteredPortsNumbers" source:source];
+    numbersLayer.textColor = [NSExpression expressionForConstantValue:[UIColor whiteColor]];
+    numbersLayer.textFontSize = [NSExpression expressionForConstantValue:@(self.icon.size.width / 2)];
+    numbersLayer.iconAllowsOverlap = [NSExpression expressionForConstantValue:@(YES)];
+    
+    // Style clusters
```
captainbarbosa: The `cloud` image got me confused with the `circle` image (since they're the same color and look similar). Perhaps replacing the cloud with a triangle or some other distinctive shape would help make this more clear?

Examples/ObjectiveC/ClusteringWithImagesExample.m
riastrad: Good point 👍. I think I like the idea of hardcoding.

Examples/ObjectiveC/ClusteringWithImagesExample.m
riastrad: Just want to note here that because this example is using `MGLSymbolStyleLayer` for the clustering, only 2  different style layers are necessary: one for labels and images and one for unclustered images. 

You need three layers if you're using a CircleStyleLayer (which doesn't allow you to set a `.text` _and_ an `.iconImageName`.

riastrad:
@captainbarbosa I think I've addressed all of your comments. Let me know if there's anything more I should tweak before merging!

riastrad:
Removed the `Podfile.lock` that was trying to sneak in via cb1cf01 and rebased.

PR #204, opened 2018-08-10
Include example Settings.bundle in project

captainbarbosa:
Move the example Settings.bundle from the Mapbox GL Native project into this project, per https://github.com/mapbox/mapbox-gl-native/issues/12554


captainbarbosa:
Scratching this, we're going to add the Settings.bundle as a resource to the `iosapp` in Mapbox GL Native instead per https://github.com/mapbox/mapbox-gl-native/issues/12554#issuecomment-412244744.

PR #226, opened 2018-09-14
Update for Xcode 10

friedbunny:
Split out of #216.

/cc @julianrex @captainbarbosa

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -2,9 +2,9 @@ platform :ios, '9.0'
 use_frameworks!
 
 def shared_pods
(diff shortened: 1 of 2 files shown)


friedbunny:
#228 did the access token fixes in 6451702, so we can remove it from this PR when we rebase.

friedbunny:
Updated SwiftLint to pull in https://github.com/realm/SwiftLint/issues/2231, which fixes the false-positive in ClusteringExample_Swift.

PR #227, opened 2018-09-18
Navigation SDK v0.21.0

1ec5:
Upgraded to navigation SDK v0.21.0.

/cc @friedbunny


1ec5:
Yes, I can reproduce that issue. Not sure what other impact it has.

PR #228, opened 2018-09-20
Ensure build phase inserts access token

1ec5:
Ensure the access token build phase takes precedence over the standard Info.plist copying step under the new build system in Xcode 10, per the updated [private access token documentation](https://www.mapbox.com/help/ios-private-access-token/).

/cc @captainbarbosa @friedbunny


friedbunny:
This supersedes part of https://github.com/mapbox/ios-sdk-examples/pull/226.

PR #231, opened 2018-10-01
Navigation SDK v0.23.0

1ec5:
Upgraded to navigation SDK ~~v0.22.0~~ v0.23.0, which changes an initializer signature in a way that only affects the Objective-C setup example.

/cc @JThramer @akitchen @friedbunny


friedbunny:
@akitchen @JThramer what’s the review status on this?

akitchen:
The change seems simple enough, but the CI task is flaky. what to do?

friedbunny:
@akitchen The flappy tests are in the main maps SDK project and are failing before the navigation target can run, so my suggestion is to rebase this PR onto `master` (now that #237 has landed) and go from there. 🐦 🙇

1ec5:
Rebased and updated to v0.23.0.

PR #248, opened 2018-11-09
ImageSourceExample: make objc/swift consistent; simplify

friedbunny:
Just some minor edits to `ImageSourceExample` and `SatelliteStyleExample` that occurred to me as I was researching another issue.

- Updates comments and keeps them consistent across objc/swift.
- Simplifies image loading.
- Removes some very deep indentation.

/cc @jmkiley @captainbarbosa


friedbunny:
For whatever reason Bitrise didn’t kick off a build when I originally pushed this, so I’ve edited the commit message and pushed again.

PR #249, opened 2018-11-15
Update Maps SDK 4.7.0

friedbunny:
Ongoing branch/PR with updates for Maps SDK 4.7.0.

/cc @mapbox/maps-ios


friedbunny:
@julianrex Now that we push pre-releases to CocoaPods’s trunk (https://github.com/mapbox/mapbox-gl-native/issues/6235), we can stick the nicer/safer standard pod versioning (`~>` and such) instead of directly specifying a podspec URL. (I pushed a change that does this.)

PR #270, opened 2019-01-31
Update SelectFeatureExample to use a more compatible layer

### LINKED ISSUE #263: SelectFeatureExample stopped working with v4.8.0-alpha.1
The example `SelectFeatureExample` on both objc and swift with the SDK version `4.8.0-alpha.1` stopped working.

jmkiley:
Fixes #263 

cc @fabian-guerra

### DIFF
--- a/Examples/Examples.m
+++ b/Examples/Examples.m
@@ -63,7 +63,7 @@
                     @{@"className": MBXExamplePointConversion, @"title": @"Point conversion"},
                     @{@"className": MBXExampleShowHideLayer, @"title": @"Show and hide a layer"},
                     @{@"className": MBXExampleFeatureSelection, @"title": @"Select a feature within a layer"},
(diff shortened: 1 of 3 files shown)


jmkiley:
@friedbunny I think it's a little more complicated to support both with this example. There is no longer a `parks` layer. Instead, there is a `landuse_overlay` that is styled based on the `class` attribute value (in this case we'd want `park`). 

We could update the example to use Streets v11 if we want to avoid pinning examples to specific style versions. If that is the case, there are [other examples](https://github.com/mapbox/ios-sdk-examples/search?q=withversion&unscoped_q=withversion) that should also be updated.

jmkiley:
Sounds good. I can change it to use water.

jmkiley:
Updated the example to use water features.

PR #274, opened 2019-02-23
Conditionally install Maps SDK builds

captainbarbosa:
This is in preparation for bringing this project as a submodule into Mapbox GL Native.

In this PR, the Podfile's `shared_pods` block checks if it is part of the Mapbox GL Native project. If it is, then it will install a local build of the Maps SDK. If not, then it installs the current released version of the Maps SDK.


Podfile
```diff
@@ -2,7 +2,13 @@ platform :ios, '9.0'
 use_frameworks!
 
 def shared_pods
-    pod 'Mapbox-iOS-SDK', '~> 4.8.0'
+    # This uses a local build of the Maps SDK if you are installing from within mapbox-gl-native
+    if Dir.pwd.include?('mapbox-gl-native')
+      pod 'Mapbox-iOS-SDK', :path => '../../../build/ios/pkg/dynamic/Mapbox-iOS-SDK.podspec'
```
friedbunny: We should note that this file/build won’t exist unless the developer does a `make iframework`.

PR #286, opened 2019-04-02
[ios] Add example for label placement.

fabian-guerra:
Example for label collision placement.

PR #285, opened 2019-04-02
[ios] Add example for formatting expressions.

fabian-guerra:
WIP requires beta.2 to work.

PR #289, opened 2019-04-04
Add missing icons example.

fabian-guerra:
Adds a new example for missing icons event.


fabian-guerra:
This is how it should look
missing-icons

riastrad:
Thanks for putting in the legwork to add the new example, @fabian-guerra! 

Would you mind sharing a quick blurb about what types of situations this usage addresses? I ask because if we add it to the docs.mapbox.com/ios/maps/examples in the future, it will be helpful to know what keywords users will be ⌘+F'ing the page for.

captainbarbosa:
@fabian-guerra do you think you can just address https://github.com/mapbox/ios-sdk-examples/pull/289#discussion_r272337130 before merging? I'm ok with the design/comment changes coming in a later PR but just a little concerned about the Pod changes as those aren't typically changed in other new example PRs.

fabian-guerra:
Hi, @riastrad. This "example" is more of a test case than a formal example. Clarifying the goal for this, and the other pr's I have #286 & #285. The goal for this feature is give developers a chance to load a style icon when is missing from the style —for many reasons, either it was not downloaded or the style was corrupted—

This as is will not be part of `ios-sdk` docs.

PR #278, opened 2019-03-07
Update for Maps SDK 4.10.0 (release-liquid)

jmkiley:

PR #290, opened 2019-04-04
Update for Maps SDK 4.11.0 (release-mojito)

friedbunny:
~Includes #278~, as we need that and it can’t be merged to `master` until we’re finished with that release.


friedbunny:
Rebased this on `master` to remove all of the commits for the previous version.

PR #300, opened 2019-05-04
Custom callout presentation should call delegate methods.

julianrex:
Custom `MGLCalloutView` callouts should call `calloutViewWillAppear` / `calloutViewDidAppear` - without this, callout views can be out of synch (due to the toggling of `presentsWithTransaction:` introduced in https://github.com/mapbox/mapbox-gl-native/pull/14307).


julianrex:
I do wonder though if it's implied though by the definition of the protocols (calling the will/did methods) 🤔

PR #301, opened 2019-05-07
Update for Maps SDK 5.0.0 (release-nectar) 

friedbunny:
~Currently branches from https://github.com/mapbox/ios-sdk-examples/pull/290, as we’re again doing multiple simultaneous releases — will need to be rebased and retargeted once that merges.~

Done.


julianrex:
Rebased on master (after landing v4.11.0 branch)

captainbarbosa:
Per chat with @friedbunny I went ahead and removed the DocsCode pod dependency in order to allow us to upgrade the SDK version in this PR.

Podfile
```diff
@@ -2,27 +2,22 @@ platform :ios, '9.0'
 use_frameworks!
 
 def shared_pods
-    # This uses a local build of the Maps SDK if you are installing from within mapbox-gl-native
-    # Build the framework first with `make iframework` to use this option.
-    if Dir.pwd.include?('mapbox-gl-native')
```
julianrex: Do we still need this?

friedbunny:
To make tests pass, the DocsCode target is also going to need to be disabled on Bitrise.

Podfile
captainbarbosa: Nope - it was implemented for something we're no longer going to do in GL-native. I'd rather take it out.

captainbarbosa:
Temporarily disabled that step in the Bitrise.yml 👍

Podfile
```diff
@@ -2,27 +2,22 @@ platform :ios, '9.0'
...
-    else
-      pod 'Mapbox-iOS-SDK', '4.11.0'
-    end
-
     pod 'SwiftLint', '~> 0.29'
 end
 
 target 'Examples' do
   shared_pods
+  pod 'Mapbox-iOS-SDK', '5.0.0'
```
friedbunny: This should stay in `shared_pods`, if the intention is to reenable `DocsCode` at some point.

Podfile
friedbunny: Hrm, can you link me to that discussion, @captainbarbosa?

friedbunny:
@captainbarbosa please do not squash merge these PRs — it loses valuable context and groups unrelated changes into a single commit.

PR #315, opened 2019-05-28
Enable more warnings and static analyzer

friedbunny:
- Enable static analyzer during all builds — this app doesn’t take long to compile and this will surface errors more quickly.
- Enable a variety of non-default warnings and fix the resulting issues.

/cc @julianrex


Examples/ObjectiveC/UserLocationAnnotationExample.m
```diff
@@ -33,7 +33,7 @@ - (void)update {
 
 - (void)updateHeading {
     // Show the heading arrow, if the heading of the user is available.
-    if (self.userLocation.heading.trueHeading) {
+    if (self.userLocation.heading.trueHeading > 0) {
```
julianrex: Can this go negative?

friedbunny: Correct values for [`CLHeading.trueHeading`](https://developer.apple.com/documentation/corelocation/clheading/1423568-trueheading?language=objc) are `0` to `360`, with negative meaning there is no heading available (which is what this is checking for).

PR #316, opened 2019-05-28
Update to Swift 5

friedbunny:
Migrating to Swift 5 was smooth, though requiring `@unknown default` for exhaustive `enum`s feels a bit excessive. More on that in https://github.com/mapbox/mapbox-gl-native/issues/14787.

/cc @mapbox/maps-ios

### DIFF
--- a/Examples/Swift/UserTrackingModesExample.swift
+++ b/Examples/Swift/UserTrackingModesExample.swift
@@ -4,7 +4,6 @@ import Mapbox
 
 class UserTrackingModesExample_Swift: UIViewController, MGLMapViewDelegate {
     var mapView: MGLMapView!
-    @IBOutlet var button: UserLocationButton!
 
     override func viewDidLoad() {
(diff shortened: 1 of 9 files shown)


Examples/Swift/UserTrackingModesExample.swift
```diff
@@ -4,7 +4,6 @@ import Mapbox
 
 class UserTrackingModesExample_Swift: UIViewController, MGLMapViewDelegate {
     var mapView: MGLMapView!
-    @IBOutlet var button: UserLocationButton!
```
jmkiley: Can this property also be removed from the Objective-C example? https://github.com/mapbox/ios-sdk-examples/blob/master/Examples/ObjectiveC/UserTrackingModesExample.m#L112

friedbunny: → 152c49

PR #317, opened 2019-05-30
Update for Maps SDK v5.1.0 (release-oolong)

captainbarbosa:
Working branch for QA, to be merged after the final 5.1.0 release.


1ec5:
As of mapbox/mapbox-navigation-ios#2143, the navigation SDK is incompatible with map SDK v5.1.0. Before the navigation SDK can upgrade, we need to remove two workarounds that would result in overcorrection in v5.1.0. (This situation stems from the map SDK making backwards-incompatible changes in semver-minor releases. But at least we foresaw the problem well enough in advance that the navigation SDK could block the upgrade to v5.1.0.)

friedbunny:
> the navigation SDK is incompatible with map SDK v5.1.0

Indeed! I’ve disabled the Nav SDK integration again here in 9e3356d so that this PR can merge today.

Issue #295, opened 2019-04-19
Cache Management API Example


jmkiley:
Once https://github.com/mapbox/mapbox-gl-native/pull/14365 lands, we should add an example for how to clear the maps cache manually in an iOS app. 

This example should include how to trigger cache clearing and reinitializing from the end user's end (preferably via a button). 

From my understanding, modifying the cache via remote notification updates would require the [`application(_:didReceiveRemoteNotification:fetchCompletionHandler:)`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application) method in the `AppDelegate` file, as well as permissions adjustments. That is a little more complicated than our typical examples. Therefore, we may want to implement it in the examples app itself, use `iosapp`, or create a tutorial.

jmkiley:
After chatting with @captainbarbosa, I think that this should be a guide rather than an example. This allows us to give context for when this API should be used, plus keep multiple possible implementations in one place.

cc @langsmith @zugaldia

jmkiley:
Landed in #323.

PR #323, opened 2019-07-17
Add cache management API example

jmkiley:
Addresses #295 

Examples in progress. I was hoping to show the `cache.db` size before and after clearing the ambient cache or resetting the database, but struggling a bit with the Swift implementation.

cc @langsmith


jmkiley:
Per chat with @captainbarbosa, I created an `Advanced` folder for this example. I don't think that it belongs on the website, but having demo code to link developers to should help those who need this functionality.

Examples.xcodeproj/project.pbxproj
```diff
@@ -130,6 +130,10 @@
 		96BAED2E1FE304DD00A58BEB /* PointConversionExample.m in Sources */ = {isa = PBXBuildFile; fileRef = 96BAED2D1FE304DD00A58BEB /* PointConversionExample.m */; };
 		96D431FC1C84B4F7007D09D1 /* ImageAnnotationExample.m in Sources */ = {isa = PBXBuildFile; fileRef = 96D431FB1C84B4F7007D09D1 /* ImageAnnotationExample.m */; };
 		96D432061C84B9CF007D09D1 /* pisavector.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 96D432051C84B9CF007D09D1 /* pisavector.xcassets */; };
+		A42F4A4D22EF9A79005097F3 /* CacheManagementExample.m in Sources */ = {isa = PBXBuildFile; fileRef = A42F4A4C22EF9A79005097F3 /* CacheManagementExample.m */; };
+		A42F4A5122EF9ACD005097F3 /* CacheManagementExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = A42F4A5022EF9ACD005097F3 /* CacheManagementExample.swift */; };
+		A4D01B4E22DE5F8300595A37 /* (null) in Sources */ = {isa = PBXBuildFile; };
+		A4D01B5122DE623700595A37 /* (null) in Sources */ = {isa = PBXBuildFile; };
```
friedbunny: 🤔

Examples/AppDelegate.m
```diff
@@ -20,6 +21,13 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
         self.window.layer.speed = 100;
     }
 
+    NSUInteger maxCacheSize = 62914560;
+    [[MGLOfflineStorage sharedOfflineStorage] setMaximumAmbientCacheSize:maxCacheSize withCompletionHandler:^(NSError * _Nullable error) {
+        if (error != nil) {
+            NSLog(@"Unable to set maximum ambient cache size");
```
friedbunny: Would it be useful to also print the error or its description?

Examples/AppDelegate.m
```diff
@@ -20,6 +21,13 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
         self.window.layer.speed = 100;
     }
 
+    NSUInteger maxCacheSize = 62914560;
+    [[MGLOfflineStorage sharedOfflineStorage] setMaximumAmbientCacheSize:maxCacheSize withCompletionHandler:^(NSError * _Nullable error) {
```
friedbunny: Hrm, can you explain why we’re doing this globally in the app delegate?

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,149 @@
...
+    self.mapView.delegate = self;
+    [self.view addSubview:self.mapView];
+
+    // Create an offline pack.
+    [self addOfflinePack];
+
+    // Add a bar button. Tapping this button will present a menu of options. For this example, the cache is managed through the UI. It can also be managed by developers through remote notifications.
+    // For more information about managing remote notifications in your iOS app, see the Apple "UserNotifications" documentation: https://developer.apple.com/documentation/usernotifications?language=objc
+    UIBarButtonItem *alertButton = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemOrganize target:self action:@selector(presentActionSheet)];
+    [self.navigationController.navigationItem setRightBarButtonItem:alertButton];
```
friedbunny: The navigation controller actually belongs to the parent view controller here, which is why this line doesn’t make the button appear on the nav bar.

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,149 @@
...
+         }]];
+
+           [self.alertController addAction:[UIAlertAction actionWithTitle:@"Reset Database" style:UIAlertActionStyleDefault handler:^(UIAlertAction * _Nonnull action) {
+              [self resetDatabase];
+          }]];
+
+        [self.alertController addAction:[UIAlertAction actionWithTitle:@"Cancel" style:UIAlertActionStyleCancel handler:nil]];
+    }
+
+    [self presentViewController:self.alertController animated:YES completion:nil];
```
friedbunny: Check this functionality on iPad — action sheets aren’t modal on larger screen sizes, but instead are presented as popovers from a particular origin.

Examples/ObjectiveC/CacheManagementExample.m
friedbunny: You may also be able to skip the `navigationController` part entirely by using `parent.navigationItem`.

Examples/AppDelegate.m
jmkiley: I put that in AppDelegate initially in order to have it happen before any map view is loaded, but I think I can move it to the actual example.

Examples.xcodeproj/project.pbxproj
jmkiley: 🤔

Examples/ObjectiveC/CacheManagementExample.m
jmkiley: Hm, so I'm running into an issue where the bar button appears on iPhones, but not on the iPad simulator.

jmkiley: Okay, I think I've addressed this.

jmkiley:
Ready for re-review.

Examples/AppDelegate.m
```diff
@@ -7,6 +7,7 @@
 //
 
 #import "AppDelegate.h"
+@import Mapbox;
```
friedbunny: This is unused.

Examples.xcodeproj/project.pbxproj
```diff
@@ -1468,9 +1494,11 @@
 				055ABCC31EFB14AE0063BACA /* PolygonPatternExample.swift in Sources */,
 				646B62AE1DEF7155000AA523 /* LineStyleLayerExample.swift in Sources */,
 				3EBCD71F1DC28240001E342F /* DefaultStylesExample.swift in Sources */,
+				A4D01B5122DE623700595A37 /* (null) in Sources */,
```
friedbunny: 🤔

Examples/AppDelegate.m
```diff
@@ -19,7 +20,6 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     if ([[NSProcessInfo processInfo].arguments containsObject:@"useFastAnimations"]) {
         self.window.layer.speed = 100;
     }
-
```
friedbunny: Nit: if we don’t need to substantively change this file, let’s also remove the white space change.

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,158 @@
+#import "CacheManagementExample.h"
+@import Mapbox;
+
+NSString *const MBXExampleCacheManagement = @"CacheManagementExample";
+
+@interface CacheManagementExample () <MGLMapViewDelegate>
+
+@property (nonatomic) MGLMapView *mapView;
+@property (nonatomic) UIAlertController *alertController;
```
friedbunny: Does this need to be a property? If it’s only used to check if something is already being presented, there may be a suitable property on the view controller already.

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,158 @@
...
+
+@implementation CacheManagementExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    /* Set the maximum ambient cache size in bytes. Call this method before the map view is loaded.
+
+     The ambient cache is created through the end user loading and using a map view. */
+    NSUInteger maxCacheSize = 62914560;
```
friedbunny: Add a comment or rename this variable to a human-readable size.

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,158 @@
...
+    [[MGLOfflineStorage sharedOfflineStorage] setMaximumAmbientCacheSize:maxCacheSize withCompletionHandler:^(NSError * _Nullable error) {
+        if (error != nil) {
+            NSLog(@"Unable to set maximum ambient cache size: %@", error.localizedDescription);
+        }
+
+        self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds];
```
friedbunny: Creating the map view inside this block strikes me as precarious/unusual — it forces you to unnecessarily consider threading and object retention.

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,158 @@
...
+        [self addOfflinePack];
+    }];
+
+    // Add a bar button. Tapping this button will present a menu of options. For this example, the cache is managed through the UI. It can also be managed by developers through remote notifications.
+    // For more information about managing remote notifications in your iOS app, see the Apple "UserNotifications" documentation: https://developer.apple.com/documentation/usernotifications?language=objc
+    UIBarButtonItem *alertButton = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemEdit target:self action:@selector(presentActionSheet)];
+    [self.parentViewController.navigationItem setRightBarButtonItem:alertButton];
+}
+
+#pragma mark: Cache management methods called by action sheet
```
friedbunny: Nit: does this colon work here?

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,158 @@
...
+    CFTimeInterval start = CACurrentMediaTime();
+    [[MGLOfflineStorage sharedOfflineStorage] invalidateAmbientCacheWithCompletionHandler:^(NSError * _Nullable error) {
+        if (error) {
+            NSLog(@"Error: %@", error.localizedDescription);
+            return;
+        } else {
+            CFTimeInterval difference = CACurrentMediaTime() - start;
+
+            // Display an alert to indicate that the invalidation is complete.
+            [self presentCompletionAlertWithTitle:@"Ambient Cache Invalidated" andMessage: [NSString stringWithFormat: @"Invalidated ambient cache in %f seconds", difference]];
```
friedbunny: References to `self` inside blocks should typically be `weak` or `unowned`. Additionally, UI work must happen on the main thread, so does this?

Examples/ObjectiveC/CacheManagementExample.m
jmkiley: 👍  I need to call the method before the map is loaded. `setMaximumAmbientCacheSize` is called on the main thread, but I can move the method back to AppDelegate.

friedbunny: Let’s keep the cache stuff contained to this example and perhaps make a note that the method might be better called elsewhere.

My suggestion for how to do the map init would be to extract it into a different method.

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,166 @@
...
+
+@implementation CacheManagementExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    /* Set the maximum ambient cache size in bytes. Call this method before the map view is loaded.
+
+     The ambient cache is created through the end user loading and using a map view. */
+    NSUInteger maximumCacheSizeInBytes = 62914560;
```
friedbunny: Sorry that I wasn’t more clear in my earlier review — the comment above this already states that it’s in bytes, but I don’t want to have to do the conversion of bytes-to-megabytes myself every time I read this or want to edit it.

Two concrete suggestions:
- Use math instead, which offers discrete units that are easier to read/edit.
- Directly comment after the definition.

```suggestion
    NSUInteger maximumCacheSizeInBytes = 64 * 1024 * 1024; // 64 megabytes
```

jmkiley:
Thanks for the feedback, @friedbunny! I think I addressed all of your comments.

Examples/ObjectiveC/CacheManagementExample.m
```diff
@@ -0,0 +1,171 @@
+#import "CacheManagementExample.h"
+@import Mapbox;
+
+NSString *const MBXExampleCacheManagement = @"CacheManagementExample";
+
+@interface CacheManagementExample () <MGLMapViewDelegate>
```
friedbunny: I might be missing something, but I don’t see that this needs to conform to `MGLMapViewDelegate`.

Examples/Swift/CacheManagementExample.swift
```diff
@@ -0,0 +1,158 @@
...
+        alertController.addAction(UIAlertAction(title: "Clear Ambient Cache", style: .default, handler: { (action) in
+            self.clearAmbientCache()
+        }))
+        alertController.addAction(UIAlertAction(title: "Reset Database", style: .default, handler: { (action) in
+            self.resetDatabase()
+        }))
+        alertController.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: nil))
+
+        alertController.popoverPresentationController?
+            .sourceView = mapView
```
friedbunny: Nit: combine this into a single line.

jmkiley:
🙇‍♀  thanks @friedbunny

jmkiley:
@friedbunny Is this okay to merge?

Podfile.lock
```diff
@@ -17,4 +17,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: e5379fa8b800bbe770f6db5b7850d7a8527ff27c
 
-COCOAPODS: 1.7.5
+COCOAPODS: 1.7.4
```
friedbunny: Let’s not downgrade CocoaPods.

```suggestion
COCOAPODS: 1.7.5
```

Issue #242, opened 2018-10-22
Animated Line Example not working in ObjC


fabian-guerra:
In the Mapbox Maps SDK for iOS v4.6.0-alpha.1 animating a line example is not working (does not display the animation nor the line). 
It works OK in swift.

julianrex:
FYI I also saw this in v4.5.0-beta.1

captainbarbosa:
Not seeing this in 4.7.0-beta.2, closing.

julianrex:
Spotted again: https://github.com/mapbox/gl-internal/issues/1055#issuecomment-458697380

fabian-guerra:
I'm not longer seeing this issue (😅). If I spotted again I will fix it right away.

Closing this ticket meanwhile.

captainbarbosa:
This seems to be happening intermittently - I was able to reproduce it once but when I restarted the app it didn't occur. Reopening for now.

julianrex:
This is still here. I suspect (though have not reproduced) that `mapViewDidFinishLoadingMap` is not getting called.

Wondering if there is any connection to https://github.com/mapbox/mapbox-gl-native/issues/13867.

Or perhaps overlapping examples/requests sharing the same cache.

jmkiley:
This should be fixed with #331.

julianrex:
I’m not sure this is true :(

julianrex:
Still happening in 5.7.0

PR #331, opened 2019-08-06
Fix timer not firing in AnimatedLineExample.

fabian-guerra:
Fixes https://github.com/mapbox/ios-sdk-examples/issues/242.


Examples/ObjectiveC/AnimatedLineExample.m
```diff
@@ -9,6 +9,7 @@ @interface AnimatedLineExample () <MGLMapViewDelegate>
 @property (nonatomic) MGLShapeSource *polylineSource;
 @property (nonatomic) NSArray<CLLocation *> *locations;
 @property (nonatomic) NSInteger currentIndex;
+@property (nonatomic, strong) NSTimer *timer;
```
friedbunny: `strong` is implicit for these property declarations in ARC code, but I suppose being explicit here is also fine.

PR #342, opened 2019-08-21
OfflinePackExample: limit impact of offline pack downloads on the app

### LINKED ISSUE #340: OfflinePackExample should not download automatically
`OfflinePackExample` currently kicks off a ~1000 tile offline region download when it loads — this should probably be:

- Converted to use a button.
- Disabled during generic UI tests.
- Made to stop downloading the offline region after the view controller has disappeared.

/cc @captainbarbosa

friedbunny:
Fixes #340. Should let UI tests be marginally better performant/more reliable, since offline pack downloads won’t spill over outside the relevant view controller.

/cc @captainbarbosa @jmkiley

### DIFF
--- a/Examples/ObjectiveC/OfflinePackExample.m
+++ b/Examples/ObjectiveC/OfflinePackExample.m
@@ -34,8 +34,20 @@ NSString *const MBXExampleOfflinePack = @"OfflinePackExample";
     [self startOfflinePackDownload];
 }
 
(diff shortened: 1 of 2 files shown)


friedbunny:
> 👍 Don't forget to PR these changes to the docs 

This will hopefully happen as part of the `ios-v5.3.0` release process, as this is merging into that branch.

PR #349, opened 2019-09-10
[ios] Fixes an issue that caused a timer run in a background thread.

fabian-guerra:
Fixes an issue that caused a random execution for the animated line example.


Examples/ObjectiveC/AnimatedLineExample.m
```diff
@@ -59,7 +59,8 @@ - (void)animatePolyline {
     self.currentIndex = 1;
 
     // Start a timer that will simulate adding points to our polyline. This could also represent coordinates being added to our polyline from another source, such as a CLLocationManagerDelegate.
-    self.timer = [NSTimer scheduledTimerWithTimeInterval:0.05 target:self selector:@selector(tick:) userInfo:nil repeats:YES];
```
friedbunny: So, [the docs](https://developer.apple.com/documentation/foundation/nstimer) say that `+[NSTimer scheduledTimerWithTimeInterval:target:selector:userInfo:repeats:]` “create[s] the timer and schedule[s] it on the current run loop in the default mode.”

Looking at this example’s view controller, this timer should have already been called on the main thread and main run loop — there don’t appear to be any `dispatch` calls or other threading methods.

That would make the difference in this PR just that we now use [`NSRunLoopCommonModes`](https://developer.apple.com/documentation/foundation/nsrunloopcommonmodes?language=objc) (and not [`NSDefaultRunLoopMode`](https://developer.apple.com/documentation/foundation/nsdefaultrunloopmode)), which looks like it will improve how often the timer fires (e.g., during input tracking), but I’m not we’ve found/solved whatever the underlying problem is.

friedbunny: From which thread/loop is `mapViewDidFinishLoadingMap` called? Presumably the main one?

PR #353, opened 2019-09-18
🧹 Clean up MultipleShapesExample

captainbarbosa:
Now shows three shapes (point, line, polygon), cleans up GeoJSON, and makes things Swiftier.

### DIFF
--- a/Examples/ObjectiveC/MultipleShapesExample.m
+++ b/Examples/ObjectiveC/MultipleShapesExample.m
@@ -7,15 +7,20 @@ NSString *const MBXExampleMultipleShapes = @"MultipleShapesExample";
 @property (nonatomic) MGLMapView *mapView;
 @end
 
+@interface MGLStyle (LayerExtensions)
+- (void)addLinesFrom:(MGLShapeSource *)source;
+- (void)addPointsFrom:(MGLShapeSource *)source;
+- (void)addPolygonsFrom:(MGLShapeSource *)source;
+@end
+
 @implementation MultipleShapesExample
 
 - (void)viewDidLoad {
     [super viewDidLoad];
-    
+
+    // Configure a map view centered on Washington, D.C.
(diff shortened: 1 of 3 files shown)


Examples/ObjectiveC/MultipleShapesExample.m
```diff
@@ -7,61 +7,80 @@ @interface MultipleShapesExample () <MGLMapViewDelegate>
...
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        NSString *path = [[NSBundle mainBundle] pathForResource:@"metro-line" ofType:@"geojson"];
-        
-        NSData *data = [NSData dataWithContentsOfFile:path];
-        
-        dispatch_async(dispatch_get_main_queue(), ^{
-            [self drawShapeCollection:data];
-        });
-    });
+    // Since we know this file exists within this project, we'll force unwrap its value.
```
jmkiley: This comment doesn't seem applicable for the Objective-C example. It may be more relevant to say something like:

 > If this data were coming from an external server, we would want to check whether the `NSURL` had a value of `nil`.

Examples/ObjectiveC/MultipleShapesExample.m
```diff
@@ -7,61 +7,80 @@ @interface MultipleShapesExample () <MGLMapViewDelegate>
...
-    [self.mapView.style addSource:source];
-    
-    // Create station style layer.
+@end
+
+@implementation MGLStyle (LayerExtensions)
+
+- (void)addLinesFrom:(MGLShapeSource *)source {
+    // Configure a line style layer to represent a rail line, filtering out all data from the
+    // source that is not of `Rail line` type.
```
jmkiley: Should we say that `TYPE` is an attribute key, just so people know where `Rail line` and the other types are coming from?

PR #359, opened 2019-09-19
Add ShapeAnnotationsExample

captainbarbosa:
Adds a new example to showcase how to add three types of annotations to one map, all with callouts:

Screenshot 2019-09-19 16 28 14

This example was made in part to subsume `DefaultCalloutExample`,  `PolygonAnnotationExample`, and `LineAnnotationGeoJSONExample` (although this example doesn't add data from parsed GeoJSON, we have other examples of this already).

### DIFF
--- a/Examples/ObjectiveC/ShapeAnnotationsExample.m
+++ b/Examples/ObjectiveC/ShapeAnnotationsExample.m
@@ -1,0 +1,91 @@
+#import "ShapeAnnotationsExample.h"
+@import Mapbox;
+
+NSString *const MBXExampleShapeAnnotations = @"ShapeAnnotationsExample";
+
+@interface ShapeAnnotationsExample () <MGLMapViewDelegate>
+
+@property MGLPointAnnotation *pointAnnotation;
+@property MGLPolyline *lineAnnotation;
+@property MGLPolygon *polygonAnnotation;
+
+@end
+
+@implementation ShapeAnnotationsExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.frame];
+    mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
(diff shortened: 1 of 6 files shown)


Examples/ObjectiveC/ShapeAnnotationsExample.m
```diff
@@ -0,0 +1,85 @@
...
+
+    NSUInteger numberOfLineCoordinates = sizeof(lineCoordinates) / sizeof(CLLocationCoordinate2D);
+
+    CLLocationCoordinate2D polygonCoordinates[] = {
+        CLLocationCoordinate2DMake(45.512, -122.663),
+        CLLocationCoordinate2DMake(45.512, -122.654),
+        CLLocationCoordinate2DMake(45.522, -122.654),
+        CLLocationCoordinate2DMake(45.522, -122.663),
+        CLLocationCoordinate2DMake(45.512, -122.663)
+    };
```
friedbunny: Putting these all in `viewDidLoad` (before the map view is set up, even) feels a bit unorganized. Could these be broken out into properties/methods?

Examples/ObjectiveC/ShapeAnnotationsExample.m
```diff
@@ -0,0 +1,96 @@
...
+    mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    [mapView setCenterCoordinate:CLLocationCoordinate2DMake(45.520, -122.663) zoomLevel:13 animated:NO];
+    // Allow this ViewController class to be the reciever of `MGLMapViewDelegate` events.
+    mapView.delegate = self;
+    [self.view addSubview:mapView];
+
+    [self setupCoordinates];
+
+    // Three shapes annotations are created below: a point, line, and a polygon.
+    MGLPointAnnotation *pointAnnotation = [[MGLPointAnnotation alloc] init];
```
jmkiley: It might be a bit neater to move the creation of these annotations to a separate method as well. You could probably make it work that way without needing to store the annotations as properties.

captainbarbosa:
yes

PR #362, opened 2019-09-23
Remove DefaultAnnotationExample

captainbarbosa:
...in favor of the example added in https://github.com/mapbox/ios-sdk-examples/pull/359, which shows not only how to display a callout with just the default point annotation but also line and polygon annotations.

Must merge _after_ https://github.com/mapbox/ios-sdk-examples/pull/359.


captainbarbosa:
I got this confused with `DefaultAnnotationExample`, which doesn't use accessory views. The PR has been updated to reflect this. We should definitely keep `DefaultCalloutExample`, and rename it to something more descriptive of the purpose it serves.

PR #376, opened 2019-11-19
Change slider configuration to non-continuous in order to suppress change events.

### LINKED ISSUE #375: UISlider update events result in compute-intensive apis being called on main thread, causing laggy behavior and occasional crash.
In the `BuildingLightExample` and `RasterImageryExample` (and their swift counterparts), a `UISlider` is used to allow the user to play with various effects. This slider however is currently configured to be `continuous` and as a result invokes it's selector (on main thread) on every update event. 

Since the mapbox operations are compute-intensive this causes laggy behavior in simulators (and the occasional crash on Xcode).

nishant-karajgikar:
Resolves #375 by suppressing change events.


nishant-karajgikar:
Thanks for the review!

PR #382, opened 2019-12-11
Add change worldview boundaries example

ZiZasaurus:
created new iOS example for changing worldview boundaries

### DIFF
--- a/Examples/ObjectiveC/ChangeWorldviewBoundaries.m
+++ b/Examples/ObjectiveC/ChangeWorldviewBoundaries.m
@@ -1,0 +1,95 @@
+#import "ChangeWorldviewBoundaries.h"
+
+@import Mapbox;
+
+NSString *const MBXExampleChangeWorldviewBoundaries = @"ChangeWorldviewBoundariesExample";
+
+@interface ChangeWorldviewBoundariesExample ()
+
+@property (nonatomic) MGLMapView *mapView;
+@property MGLLineStyleLayer *layer;
+
+
+@end
+
+@implementation ChangeWorldviewBoundariesExample
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    self.mapView = [[MGLMapView alloc] initWithFrame: CGRectMake(0, 0, self.view.bounds.size.width, self.view.bounds.size.height)];
+    self.mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    
+    // Set the map's initial style, center coordinate, and zoom level
+    self.mapView.styleURL = [MGLStyle lightStyleURL];
+    [self.mapView setCenterCoordinate:CLLocationCoordinate2DMake(25.251, 95.69)
+                            zoomLevel:3
+                             animated:NO];
+    [self.view addSubview:self.mapView];
+    
+    // Create a UISegmentedControl to toggle between map styles
+    UISegmentedControl *styleToggle =[[UISegmentedControl alloc] initWithItems:@[@"  US  ", @"  CN  ", @"  IN  ", @"  All  "]];
+    styleToggle.translatesAutoresizingMaskIntoConstraints = NO;
+    styleToggle.tintColor = [UIColor colorWithRed:0.976f green:0.843f blue:0.831f alpha:1];
+    styleToggle.backgroundColor = [UIColor colorWithRed:0.721f green:0.804f blue:0.831f alpha:1];
+    styleToggle.layer.cornerRadius = 4;
+    styleToggle.clipsToBounds = YES;
+    [self.view insertSubview:styleToggle aboveSubview:self.mapView];
+    [styleToggle addTarget:self action:@selector(changeStyle:) forControlEvents:UIControlEventValueChanged];
+    
+    // Configure autolayout constraints for the UISegmentedControl to align
+    // at the bottom of the map view and above the Mapbox logo and attribution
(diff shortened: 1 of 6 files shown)


Examples/Swift/ChangeWorldviewBoundariesExample.swift
```diff
@@ -0,0 +1,76 @@
+import UIKit
+import Mapbox
+
+@objc(ChangeWorldviewBoundariesExample_Swift)
+
+class ChangeWorldviewBoundariesExample: UIViewController, MGLMapViewDelegate {
+
+    let mapView = MGLMapView()
+    let source = MGLVectorTileSource(identifier: "admin-bounds", configurationURL: NSURL(string: "mapbox://mapbox.mapbox-streets-v8")! as URL)
```
captainbarbosa: If this source is already included in Mapbox Light by default, you can access it instead with `MGLStyle.source(identifier: String)`, which returns an optional `MGLSource`, instead of declaring a new source.

Examples/Swift/ChangeWorldviewBoundariesExample.swift
```diff
@@ -0,0 +1,76 @@
...
+    var adminLayers = [String]()
```
captainbarbosa: `adminLayers` is declared here but not used.

Examples/Swift/ChangeWorldviewBoundariesExample.swift
```diff
@@ -0,0 +1,76 @@
...
+    var layer: MGLLineStyleLayer?
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        mapView.frame = CGRect(x: 0, y: 0, width: view.bounds.width, height: view.bounds.height)
```
captainbarbosa: You can assign this to `self.view.frame` instead of using the entire `CGRect` constructor.

Examples/Swift/ChangeWorldviewBoundariesExample.swift
```diff
@@ -0,0 +1,76 @@
+import UIKit
+import Mapbox
+
+@objc(ChangeWorldviewBoundariesExample_Swift)
+
+class ChangeWorldviewBoundariesExample: UIViewController, MGLMapViewDelegate {
```
captainbarbosa: The view controller is conforming to the `MGLMapViewDelegate` protocol but is not using any of its associated delegate methods (yet).

Examples/Swift/ChangeWorldviewBoundariesExample.swift
```diff
@@ -0,0 +1,76 @@
...
+
+        guard let style = self.mapView.style else {return}
+
+        if layer?.identifier == "admin" {
+            print("source already exists")
+            style.removeLayer(layer!)
+        } else {
+            layer = MGLLineStyleLayer(identifier: "admin", source: source)
+            style.addSource(source)
+        }
```
captainbarbosa:
```swift
layer = MGLLineStyleLayer(identifier: "admin", source: source)
style.addSource(source)
```

This only happens the first time a user selects a segment. I don't think this code is necessary, since it appears that Mapbox Light already includes admin boundary style layers. That layer just needs to be accessed with `MGLStyle.layer(withIdentifier: String)`, so we can avoid creating a new source and style layer for this.

Examples/Swift/ChangeWorldviewBoundariesExample.swift
```diff
@@ -0,0 +1,76 @@
...
+        styleToggle.translatesAutoresizingMaskIntoConstraints = false
+        styleToggle.tintColor = UIColor(red: 0.976, green: 0.843, blue: 0.831, alpha: 1)
+        styleToggle.backgroundColor = UIColor(red: 184/255, green: 205/255, blue: 212/255, alpha: 1)
+        styleToggle.layer.cornerRadius = 4
+        styleToggle.clipsToBounds = true
+//        styleToggle.selectedSegmentIndex = 0
+        view.insertSubview(styleToggle, aboveSubview: mapView)
+        styleToggle.addTarget(self, action: #selector(changeStyle(sender:)), for: .valueChanged)
+
+        // Configure autolayout constraints for the UISegmentedControl to align
```
captainbarbosa: `autolayout` -> `Auto Layout`

Examples/Swift/ChangeWorldviewBoundariesExample.swift
```diff
@@ -0,0 +1,76 @@
...
+            style.addSource(source)
+        }
+        layer?.sourceLayerIdentifier = "admin"
+        layer?.lineColor = NSExpression(forConstantValue: UIColor.lightGray)
+
+        switch sender.selectedSegmentIndex {
+        case 0:
+
+           layer?.predicate = NSPredicate(format: "CAST(worldview, 'NSString') == 'US'")
+           style.addLayer(layer!)
```
captainbarbosa: Once you refactor this to use the built-in admin layer, you can remove all instanced of `addLayer` in this switch statement.

Examples/ObjectiveC/ChangeWorldviewBoundaries.m
```diff
@@ -0,0 +1,95 @@
...
+
+@import Mapbox;
+
+NSString *const MBXExampleChangeWorldviewBoundaries = @"ChangeWorldviewBoundariesExample";
+
+@interface ChangeWorldviewBoundariesExample ()
+
+@property (nonatomic) MGLMapView *mapView;
+@property MGLLineStyleLayer *layer;
+
```
jmkiley:
```suggestion

```

PR #392, opened 2020-03-18
Maps SDK v5.8.0

jmkiley:
This PR is for testing v5.8.0.


1ec5:
I’ve updated StaticSnapshotExample to avoid capturing the snapshotter in the completion handler, which is no longer necessary as of mapbox/mapbox-gl-native-ios#211.

nishant-karajgikar:
@jmkiley any idea why circleCI is failing here?

jmkiley:
@nishant-karajgikar It looks like we still have a CircleCI webhook on Github, so I'll remove that. Thanks!

PR #397, opened 2020-05-30
Add InsetMapExample

ZiZasaurus:
Added an iOS example similar to https://docs.mapbox.com/android/maps/examples/inset-map/


Examples/ObjectiveC/InsetMapExample.m
```diff
@@ -0,0 +1,71 @@
...
+    [self.miniMapView.layer setBorderWidth:1];
+    
+    [self.miniMapView setCenterCoordinate:self.mapView.centerCoordinate zoomLevel:self.mapView.zoomLevel animated:NO];
+    [self.miniMapView setTranslatesAutoresizingMaskIntoConstraints:NO];
+
+    // Set the delegate property of our map view to self after instantiating it
+    self.mapView.delegate = self;
+
+    [self.view addSubview:self.mapView];
+    [self.view addSubview:_miniMapView];
```
captainbarbosa, line 42: Please use the `self.property` syntax instead of `_property`. This is more consistent with the rest of our examples. If you're curious as to what the difference is, here is the [short explanation](https://teamtreehouse.com/community/objc-underscore-self-notation-difference#:~:text=When%20you%20have%20an%20object,calling%20the%20instance%20variable%20directly.) and here's the [long explanation](https://useyourloaf.com/blog/understanding-your-objective-c-self/).

Examples/ObjectiveC/InsetMapExample.m
```diff
@@ -0,0 +1,71 @@
...
+    [self installConstraints];
+
+    
+}
+
+- (void)mapView:(MGLMapView *)mapView regionWillChangeAnimated:(BOOL)animated {
+    [self.miniMapView setCamera:self.mapView.camera animated:NO
+     ];}
```
captainbarbosa, line 50: Nit: Extra spacing before `];}` - also the `;` doesn't need to be there. The end of the array should be placed on line 49 as well.

Examples/Swift/InsetMapExample.swift
```diff
@@ -0,0 +1,57 @@
+import Mapbox
+
+@objc(InsetMapExample_Swift)
+
+class InsetMapExample_Swift: UIViewController, MGLMapViewDelegate {
+    var mapView = MGLMapView()
```
captainbarbosa, line 6: For consistency, these map view variables should be declared as `var mapView: MGLMapView!`

Force unwrapping with `!` isn't great from a production standpoint, but I'd rather us be consistent.

Examples/Swift/InsetMapExample.swift
```diff
@@ -0,0 +1,57 @@
...
+        super.viewDidLoad()
+
+        mapView = MGLMapView(frame: view.bounds)
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.delegate = self
+        
+        // Set the main map's center coordinate and zoom level.
+        mapView.setCenter(CLLocationCoordinate2D(latitude: 59.31, longitude: 18.06), zoomLevel: 9, animated: false)
+
+        // set inset mapview properties
```
captainbarbosa, line 19: We should explain why we're doing this - what behavior does this result in? Its totally fine for examples to be verbose in comments.

Examples/Swift/InsetMapExample.swift
```diff
@@ -0,0 +1,57 @@
...
+        miniMapview.allowsZooming = false
+        miniMapview.allowsRotating = false
+        miniMapview.compassView.isHidden = true
+        miniMapview.logoView.isHidden = true
+        miniMapview.attributionButton.tintColor = UIColor.clear
+        let black = UIColor.black
+        miniMapview.layer.borderColor = black.cgColor
+        miniMapview.layer.borderWidth = 1
+        
+        // set inset mapview's cen
```
captainbarbosa, line 31: Typo - also for consistency please capitalize/sentence-case all comments.

Examples/Swift/InsetMapExample.swift
```diff
@@ -0,0 +1,57 @@
...
+        miniMapview.translatesAutoresizingMaskIntoConstraints = false
+        
+        view.addSubview(mapView)
+        mapView.addSubview(miniMapview)
+        installConstraints()
+       
+    }
+ 
+    func mapViewRegionIsChanging(_ mapView: MGLMapView) {
+        miniMapview.setCamera(mapView.camera, animated: false)
```
captainbarbosa, line 42: Please add an explanation as to what is happening here, since it is important.

Examples/ObjectiveC/InsetMapExample.m
ZiZasaurus: xCodee won't let me remove the `;` :(

ZiZasaurus:
Unable to set `self.miniMapView.logoView.isHidden = YES`

oddly enough, the `miniMapview` for swift is zoomed in closer than the one for obj-c and I don't know why...

Otherwise, all other changes were made!

Examples/ObjectiveC/InsetMapExample.m
captainbarbosa: Ah, my bad - I just noticed the `;` is part of the ending statement of `setCamera`. You most recent change corrected this.

captainbarbosa:
>Unable to set self.miniMapView.logoView.isHidden = YES

Oh, are you getting this error? `No setter method 'setIsHidden:' for assignment to property`

I think this is expected due to the way Objective-C is handling this boolean, [check this out](https://stackoverflow.com/questions/28528839/no-setter-method-for-assignment-to-property-cocoa-application). So I think you want to use `hidden` instead of `setHidden`. In Swift, the same property is `isHidden` 🙂

Examples/ObjectiveC/InsetMapExample.m
```diff
@@ -0,0 +1,74 @@
...
+    [super viewDidLoad];
+
+    self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds];
+    self.mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+    // Set the main map's center coordinate and zoom level.
+    [self.mapView setCenterCoordinate:CLLocationCoordinate2DMake(18.1096,-77.2975)
+                zoomLevel:9
+                 animated:NO];
+
+    // Set inset mapview properties to create a smaller, non-interactive mapView that mimics the appearance of the main mapView.
```
captainbarbosa, line 23: The length of the lines for comments should cap at 80 characters. The best way to visualize that limit is by enabling the page guide in Xcode's preferences:




We limit this length because we don't want users to have to scroll horizontally when these examples end up making it to the website. Please wrap these comments around to use the multi-line comment syntax if they go over this character limit:

```
/**
 This
 is
 a
 multi
 line
 comment :)
*/
```

Examples/ObjectiveC/InsetMapExample.m
```diff
@@ -0,0 +1,71 @@
...
+                zoomLevel:9
+                 animated:NO];
+    
+    self.miniMapView = [[MGLMapView alloc] initWithFrame:CGRectZero];
+    [self.miniMapView setAllowsScrolling:NO];
+    [self.miniMapView setAllowsTilting:NO];
+    [self.miniMapView setAllowsZooming:NO];
+    [self.miniMapView setAllowsRotating:NO];
+
+    [self.miniMapView.logoView setHidden:YES];
```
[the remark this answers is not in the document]
captainbarbosa: I was wrong, for Objective-C this should just be `hidden` and not `isHidden`, which is the Swift syntax: `self.miniMapView.logoView.hidden = YES;`

Examples/ObjectiveC/InsetMapExample.m
```diff
@@ -0,0 +1,74 @@
...
+    [self.miniMapView.compassView setHidden: YES];
+    self.miniMapView.attributionButton.tintColor = UIColor.clearColor;
+    self.miniMapView.layer.borderColor = UIColor.blackColor.CGColor;
+    self.miniMapView.layer.borderWidth = 1;
+    
+    [self.miniMapView setCenterCoordinate:self.mapView.centerCoordinate zoomLevel:self.mapView.zoomLevel animated:NO];
+    [self.miniMapView setTranslatesAutoresizingMaskIntoConstraints:NO];
+
+    // Set the delegate property of our map view to self after instantiating it
```
captainbarbosa, line 41: Missing period at the end.

Examples/ObjectiveC/InsetMapExample.m
```diff
@@ -0,0 +1,74 @@
...
+}
+
+- (void)mapView:(MGLMapView *)mapView regionWillChangeAnimated:(BOOL)animated {
+    [self.miniMapView setCamera:self.mapView.camera animated:NO];
+}
+
+
+- (void)installConstraints {
+
+
```
captainbarbosa, line 58: Please limit spacing to one line break in between code. Looks like its two lines here.

Examples/Swift/InsetMapExample.swift
```diff
@@ -0,0 +1,61 @@
...
+        super.viewDidLoad()
+
+        mapView = MGLMapView(frame: view.bounds)
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.delegate = self
+        
+        // Set the main map's center coordinate and zoom level.
+        mapView.setCenter(CLLocationCoordinate2D(latitude: 18.1096, longitude: -77.2975), zoomLevel: 9, animated: false)
+
+        // Set inset mapView's center
```
captainbarbosa, line 19: Please refer to the map as a `map view` instead of `mapView` when referring to `MGLMapView` in this context.

Examples/Swift/InsetMapExample.swift
```diff
@@ -0,0 +1,61 @@
...
+        miniMapview = MGLMapView(frame: CGRect.zero)
+        
+        // Set inset mapview properties to create a smaller, non-interactive mapView that mimics the appearance of the main mapView.
+        miniMapview.allowsScrolling = false
+        miniMapview.allowsTilting = false
+        miniMapview.allowsZooming = false
+        miniMapview.allowsRotating = false
+        miniMapview.compassView.isHidden = true
+        
+        // we are only hiding the attribution within the second map view. Ordinarily, hiding the map view's attribution would be a violation of TOS, however, because the main map view still has its attribution, there is no issue here.
```
captainbarbosa, line 29: - `we` → `We`
- Please wrap this into a multi-line comment limited to 80 chars per line as mentioned earlier.
- Avoid using acronyms per our internal style guide. An edit suggestion:

```
/**
 Hiding the map view's attribution goes against our attribution requirements. However, because the main 
 map view still has attribution, hiding the attribution for the mini map view in this case is acceptable. 

 For more information, please refer to our attribution guidelines: https://docs.mapbox.com/help/how- mapbox-works/attribution/
*/
```

Podfile.lock
```diff
@@ -21,4 +21,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 9db81edec5b4deeb01978486713076bbb76db6fa
 
-COCOAPODS: 1.8.4
+COCOAPODS: 1.8.0
```
captainbarbosa, line 24: Can you drop this change from your PR? We shouldn't be changing to CocoaPods version here.

PR #401, opened 2020-07-02
Change zoom level for InsetMap example

ZiZasaurus:


Examples/ObjectiveC/InsetMapExample.m
```diff
@@ -47,7 +47,7 @@ - (void)viewDidLoad {
     self.miniMapView.layer.borderColor = UIColor.blackColor.CGColor;
     self.miniMapView.layer.borderWidth = 1;
     
-    [self.miniMapView setCenterCoordinate:self.mapView.centerCoordinate zoomLevel:self.mapView.zoomLevel animated:NO];
+    [self.miniMapView setCenterCoordinate:self.mapView.centerCoordinate zoomLevel:self.mapView.zoomLevel-2 animated:NO];
```
captainbarbosa, line 50: Can you please add a comment as to why we're doing this? (Same for the Swift version)

Examples/Swift/InsetMapExample.swift
```diff
@@ -61,8 +62,10 @@ class InsetMapExample_Swift: UIViewController, MGLMapViewDelegate {
      are made in the mini map view.
      */
     func mapViewRegionIsChanging(_ mapView: MGLMapView) {
-        miniMapview.setCamera(mapView.camera, animated: false)
-    }
+         miniMapview.setCenter(self.mapView.centerCoordinate,
+         zoomLevel: mapView.zoomLevel-2, animated: false)
```
captainbarbosa, line 66: Please add spacing between the operators, i.e. `mapView.zoomLevel - 2`

Examples/Swift/InsetMapExample.swift
```diff
@@ -46,8 +46,13 @@ class InsetMapExample_Swift: UIViewController, MGLMapViewDelegate {
         miniMapview.attributionButton.tintColor = UIColor.clear
         miniMapview.layer.borderColor = UIColor.black.cgColor
         miniMapview.layer.borderWidth = 1
+        
+        /**
+         Setting mini map view zoom level to 2 zoom levels below the main map view to showcase
+         one use case for an inset map.
```
captainbarbosa, line 52: >showcase one use case for an inset map

What is the use case? Let's dive in a little deeper here. From a user's perspective, why would it make sense for the mini inset map to show a more zoomed out map?

ZiZasaurus: It can be used either to display a broader geographical frame of reference, or display more detail within a particular area of the map.

PR #402, opened 2020-07-15
Update version to 5.10-beta.1

nishant-karajgikar:


captainbarbosa:
We should close this in favor of 6.0.0 right?

PR #418, opened 2020-10-06
Pushing example to manage downloaded offline regions

ZiZasaurus:
Created an example for parity between iOS and Android regarding managing offline regions

### DIFF
--- a/Examples/Swift/ManageOfflineRegionsExample.swift
+++ b/Examples/Swift/ManageOfflineRegionsExample.swift
@@ -132,8 +132,11 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
 
     @objc func startOfflinePackDownload() {
 
-        // Create a region that includes the current viewport and any tiles needed to view it when zoomed further in.
-        // Because tile count grows exponentially with the maximum zoom level, you should be conservative with your `toZoomLevel` setting.
+        /**
+         Create a region that includes the current map camera, to be captured
+         in an offline map. Note: Because tile count grows exponentially as zoom level
+         increases, you should be conservative with your `toZoomLevel` setting.
+         */
         let region = MGLTilePyramidOfflineRegion(styleURL: mapView.styleURL, bounds: mapView.visibleCoordinateBounds, fromZoomLevel: mapView.zoomLevel, toZoomLevel: mapView.zoomLevel + 2)
 
         // Store some data for identification purposes alongside the downloaded resources.
--- a/Examples/Swift/ManageOfflineRegionsExample.swift
+++ b/Examples/Swift/ManageOfflineRegionsExample.swift
@@ -115,7 +115,7 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate {
                 return
             }
 
-            //Begin downloading the map for offline use.
+            // Begin downloading the map for offline use.
             pack!.resume()
 
         }
--- a/Examples/Swift/ManageOfflineRegionsExample.swift
+++ b/Examples/Swift/ManageOfflineRegionsExample.swift
@@ -177,6 +177,7 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
 
             }
         }
+        // Reload the table to update the progress percentage for each offline pack.
         self.tableView.reloadData()
 
     }
--- a/Examples/Swift/ManageOfflineRegionsExample.swift
+++ b/Examples/Swift/ManageOfflineRegionsExample.swift
@@ -3,9 +3,8 @@ import Foundation
 
 @objc(ManageOfflineRegionsExample_Swift)
 
+class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, UITableViewDelegate, UITableViewDataSource {
 
-class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, UITableViewDelegate, UITableViewDataSource  {
-    
     lazy var mapView: MGLMapView = {
         let mapView = MGLMapView(frame: CGRect.zero)
         mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
@@ -17,7 +16,7 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
     lazy var downloadButton: UIButton = {
         let downloadButton = UIButton(frame: CGRect.zero)
         downloadButton.backgroundColor = UIColor.white
-        downloadButton.setTitleColor(.black, for: .normal)
+        downloadButton.setTitleColor(UIColor.systemBlue, for: .normal)
         downloadButton.setTitle("Download Region", for: .normal)
         downloadButton.addTarget(self, action: #selector(startOfflinePackDownload), for: .touchUpInside)
         downloadButton.layer.cornerRadius = view.bounds.width / 30
@@ -32,7 +31,7 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
         tableView.translatesAutoresizingMaskIntoConstraints = false
         return tableView
     }()
-    var progressPercentage: Float?
+    var cell = UITableViewCell()
     // make an extension for this and make necessary comments and push
     override func viewDidLoad() {
 
@@ -47,59 +46,43 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
 
         // Setup offline pack notification handlers.
         addObserver()
+        installConstraints()
 
-        if #available(iOS 11.0, *) {
-            installConstraints()
-        } else {
-            // Fallback on earlier versions
-        }
     }
 
     func addObserver() {
         NotificationCenter.default.addObserver(self, selector: #selector(offlinePackProgressDidChange), name: NSNotification.Name.MGLOfflinePackProgressChanged, object: nil)
     }
 
-    @available(iOS 11.0, *)
     func installConstraints() {
 
         NSLayoutConstraint.activate([
-            mapView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
-            mapView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
-            mapView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
-            mapView.heightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.heightAnchor, multiplier: 0.75),
+            mapView.topAnchor.constraint(equalTo: view.topAnchor),
+            mapView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            mapView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            mapView.heightAnchor.constraint(equalTo: view.heightAnchor, multiplier: 0.75),
             tableView.topAnchor.constraint(equalTo: mapView.bottomAnchor),
             tableView.leadingAnchor.constraint(equalTo: mapView.leadingAnchor),
             tableView.trailingAnchor.constraint(equalTo: mapView.trailingAnchor),
             tableView.heightAnchor.constraint(equalTo: mapView.heightAnchor, multiplier: 0.4),
-            downloadButton.topAnchor.constraint(equalTo: mapView.topAnchor, constant: 10),
-            downloadButton.leadingAnchor.constraint(equalTo: mapView.leadingAnchor),
+            downloadButton.topAnchor.constraint(equalTo: mapView.topAnchor, constant: 100),
+            downloadButton.leadingAnchor.constraint(equalTo: mapView.leadingAnchor, constant: 5),
             downloadButton.trailingAnchor.constraint(equalTo: mapView.leadingAnchor, constant: 150)
         ])
 
     }
 
     override func viewDidDisappear(_ animated: Bool) {
-
         super.viewDidDisappear(animated)
-        // When leaving this view controller, remove all offline downloads.
-        guard let packs = MGLOfflineStorage.shared.packs else { return }// or do something else }
-        MGLOfflineStorage.shared.removePack(packs[0]) { (error) in
-            // handle error if pack can't be removed
+
+        MGLOfflineStorage.shared.resetDatabase { (error) in
             if let error = error {
                 print(error)
             } else {
+                MGLOfflineStorage.shared.reloadPacks()
                 self.tableView.reloadData()
             }
-            //mechanisms to reload certain cells instead of entire tableview
         }
-
-    }
-
-    deinit {
-        // Remove offline pack observers.
-        print("Removing offline pack notification observers")
-        NotificationCenter.default.removeObserver(self)
-
     }
 
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
@@ -111,22 +94,27 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
         }
     }
 
-    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        var cell = tableView.dequeueReusableCell(withIdentifier: "cell", for: indexPath)
-        cell = UITableViewCell(style: .subtitle, reuseIdentifier: "cell")
-        if let packs = MGLOfflineStorage.shared.packs {
-            let completed = Float(packs[indexPath.row].progress.countOfResourcesCompleted)
-                let expected = Float(packs[indexPath.row].progress.countOfResourcesExpected)
-                let progress = completed / expected
-            cell.textLabel?.text = "Region \(indexPath.row + 1): size: \(packs[indexPath.row].progress.countOfBytesCompleted)"
-            cell.detailTextLabel?.text = "Percent completion: \(String(format: "%.2f", progress * 100))%"
-            cell.textLabel?.textAlignment = .left
+    func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
+        let label = UILabel()
+        if section == 0 {
+            label.backgroundColor = UIColor.white
+            label.textAlignment = .center
+            label.text = "No Offline Packs Saved"
+            return label.text
+        }
 
-        } else {
-            print("so...")
-            cell.textLabel?.text = "No Offline Packs Saved"
-            cell.textLabel?.textColor = .darkGray
-            cell.textLabel?.textAlignment = .center
+        return nil
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+
+        cell = UITableViewCell(style: .subtitle, reuseIdentifier: "cell")
+
+        if let packs = MGLOfflineStorage.shared.packs {
+
+            cell.textLabel?.text = "Region \(indexPath.row + 1): size: \(packs[indexPath.row].progress.countOfBytesCompleted)"
+            cell.detailTextLabel?.text = "Percent completion: \(String(format: "%.2f", packs[indexPath.row].progress.progressPercentage))%"
+            cell.textLabel?.textAlignment = .left
 
         }
 
@@ -138,32 +126,6 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
         guard let packs = MGLOfflineStorage.shared.packs else { return }
         if let selectedRegion = packs[indexPath.row].region as? MGLTilePyramidOfflineRegion {
             mapView.setVisibleCoordinateBounds(selectedRegion.bounds, animated: true)
-        }
-    }
-
-    @available(iOS 11.0, *)
-    func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
-        return UISwipeActionsConfiguration(actions: [
-            makeDeleteContextualAction(forRowAt: indexPath)
-        ])
-    }
-
-    @available(iOS 11.0, *)
-    private func makeDeleteContextualAction(forRowAt indexPath: IndexPath) -> UIContextualAction {
-        return UIContextualAction(style: .destructive, title: "Delete") { (action, swipeButtonView, completion) in
-            guard let packs = MGLOfflineStorage.shared.packs else { return }
-            MGLOfflineStorage.shared.removePack(packs[indexPath.row]) { (error) in
-                // handle error if pack can't be removed
-                if let error = error {
-                    print(error)
-                } else {
-                    self.tableView.reloadData()
-                }
-
-                //mechanisms to reload certain cells instead of entire tableview
-            }
-
-            completion(true)
         }
     }
 
@@ -180,7 +142,7 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
 
         // Create and register an offline pack with the shared offline storage object.
 
-        MGLOfflineStorage.shared.addPack(for: region, withContext: context) { [self] (pack, error) in
+        MGLOfflineStorage.shared.addPack(for: region, withContext: context) { (pack, error) in
             guard error == nil else {
                 // The pack couldn’t be created for some reason.
                 print("Error: \(error?.localizedDescription ?? "unknown error")")
@@ -202,14 +164,9 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
         // and the associated user info for the pack; in this case, `name = My Offline Pack`
         if let pack = notification.object as? MGLOfflinePack,
            let userInfo = NSKeyedUnarchiver.unarchiveObject(with: pack.context) as? [String: String] {
-            let progress = pack.progress
-            // or notification.userInfo![MGLOfflinePackProgressUserInfoKey]!.MGLOfflinePackProgressValue
-            let completedResources = progress.countOfResourcesCompleted
-            let expectedResources = progress.countOfResourcesExpected
 
-            // Calculate current progress percentage.
-
-            progressPercentage = Float(completedResources) / Float(expectedResources)
+            let completedResources = pack.progress.countOfResourcesCompleted
+            let expectedResources = pack.progress.countOfResourcesExpected
 
             if completedResources == expectedResources {
 
@@ -221,14 +178,30 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
 
                 // Otherwise, print download/verification progress.
 
-                print("Offline pack “\(userInfo["name"] ?? "unknown")” has \(completedResources) of \(expectedResources) resources — \(String(format: "%.2f", progressPercentage! * 100))%.")
+                print("Offline pack “\(userInfo["name"] ?? "unknown")” has \(completedResources) of \(expectedResources) resources — \(String(format: "%.2f", pack.progress.progressPercentage))%.")
 
             }
-
-            self.tableView.reloadData()
-
         }
+        self.tableView.reloadData()
 
     }
+
+}
+
+fileprivate extension MGLOfflinePackProgress {
+
+    var progressPercentage: Float {
+        if countOfResourcesExpected != 0 {
+            return Float((countOfResourcesCompleted / countOfResourcesExpected) * 100)
+        } else {
+            return 1.0
+        }
+    }
+//    var completedResources: Float {
+//        return Float(countOfResourcesCompleted)
+//    }
+//    var expectedResources: Float {
+//        return Float(countOfResourcesExpected)
+//    }
 
 }
--- a/Examples/Swift/ManageOfflineRegionsExample.swift
+++ b/Examples/Swift/ManageOfflineRegionsExample.swift
@@ -132,6 +132,7 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate {
         if let pack = notification.object as? MGLOfflinePack,
            let userInfo = NSKeyedUnarchiver.unarchiveObject(with: pack.context) as? [String: String] {
 
+            // At this point, the offline pack has finished downloading.
             if pack.progress.countOfResourcesCompleted == pack.progress.countOfResourcesExpected {
 
                 let byteCount = ByteCountFormatter.string(fromByteCount: Int64(pack.progress.countOfBytesCompleted), countStyle: ByteCountFormatter.CountStyle.memory)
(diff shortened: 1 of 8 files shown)


Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+        var mapView: MGLMapView!
+        var progressView: UIProgressView!
+        var downloadButton = UIButton(frame: CGRect.zero)
+        var tableView = UITableView()
+        var regionCount = 0
+        var regionArray = [MGLTilePyramidOfflineRegion]()
+        var packArray = [MGLOfflinePack]()
+        var userInfo = [String]()
+        var pack = MGLOfflineStorage.shared.packs
+        var progressPercentage: Float?
```
captainbarbosa, line 16: Instead of declaring this here, perhaps consider extending `MGLOfflinePackProgress` to have a computed property that returns the progress percentage. 

By doing this, the responsibility of calculating the progress of the pack moves from the `ViewController` class to the `MGLOfflinePackProgress` object itself. 

You don't need to add the extension to another file, just include it at the bottom of the file and make sure its marked as `fileprivate`, so it can't be accidentally used in future examples.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+
+            view.addSubview(mapView)
+            view.addSubview(tableView)
+            view.addSubview(downloadButton)
+            mapView.setCenter(CLLocationCoordinate2D(latitude: 22.27933, longitude: 114.16281),
+
+                              zoomLevel: 13, animated: false)
+
+            // Setup offline pack notification handlers.
+            NotificationCenter.default.addObserver(self, selector: #selector(offlinePackProgressDidChange), name: NSNotification.Name.MGLOfflinePackProgressChanged, object: nil)
```
captainbarbosa, line 48: Let's move this out to a function to keep things organized.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+            NotificationCenter.default.addObserver(self, selector: #selector(offlinePackDidReceiveError), name: NSNotification.Name.MGLOfflinePackError, object: nil)
+            NotificationCenter.default.addObserver(self, selector: #selector(offlinePackDidReceiveMaximumAllowedMapboxTiles), name: NSNotification.Name.MGLOfflinePackMaximumMapboxTilesReached, object: nil)
+            installConstraints()
+        }
+
+        func installConstraints() {
+
+            if #available(iOS 11.0, *) {
```
captainbarbosa, line 56: I think these constraints should be relative to the map view, then we could possible get around not having to do this `if #available` check.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+            downloadButton.backgroundColor = UIColor.white
+            downloadButton.setTitleColor(.black, for: .normal)
+            downloadButton.setTitle("Download Region", for: .normal)
+            downloadButton.addTarget(self, action: #selector(startOfflinePackDownload), for: .touchUpInside)
+            downloadButton.layer.cornerRadius = view.bounds.width / 30
+            downloadButton.translatesAutoresizingMaskIntoConstraints = false
+
+            view.addSubview(mapView)
+            view.addSubview(tableView)
+            view.addSubview(downloadButton)
```
captainbarbosa, line 42: The download button should be added to the map view, not the root view.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+            } else {
+                // Fallback on earlier versions
+            }
+
+        }
+
+        override func viewDidDisappear(_ animated: Bool) {
+
+            super.viewDidDisappear(animated)
+            // When leaving this view controller, suspend offline downloads.
```
captainbarbosa, line 71: I think we should instead delete all the offline packs, since they will continue to exists in the database forever until the app is deleted/reinstalled.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+            } else {
+                return packArray.count
+            }
+
+        }
+
+
+        func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+            let cell = tableView.dequeueReusableCell(withIdentifier: "cell", for: indexPath)
+            if MGLOfflineStorage.shared.packs?.count != nil {
```
captainbarbosa, line 100: I'd `guard` here instead:

```swift
guard let packs = MGLOfflinePacks.shared.packs { else return }

// Access to packs here
```

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+                let progress = Float(packArray[indexPath.row].progress.countOfResourcesCompleted)/Float(packArray[indexPath.row].progress.countOfResourcesExpected)
+                cell.textLabel!.text = "Region \(indexPath.row + 1): \n  Percent completion: \(String(format: "%.2f", progress * 100)) %"
```
captainbarbosa, line 102: I'd recommend against having a line break here - maybe just a dash instead? Then we don't need to explicitly set the cell height.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+                print(Float((packArray[indexPath.row].progress.countOfResourcesCompleted)/(packArray[indexPath.row].progress.countOfResourcesExpected) * 100))
+
+                cell.textLabel?.font = UIFont.systemFont(ofSize: 15)
```
captainbarbosa, line 105: Let's just use the system font size, meaning this line can just be removed.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+                cell.textLabel?.numberOfLines = 0
+                cell.textLabel?.lineBreakMode = .byWordWrapping
+                if #available(iOS 13.0, *) {
+                    cell.textLabel?.textColor = .darkGray
```
captainbarbosa, line 109: Let's keep the color the default so we can avoid doing this `if #available` check.

Examples/Swift/ManageOfflineRegionsExample.swift
captainbarbosa, line 102: `cell.textLabel!.text` is force unwrapped here, but not on the other text label properties. I think you're fine just using `cell.textLabel?.text`.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+        func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+            let selectedRegion = regionArray[indexPath.row]
+            mapView.setVisibleCoordinateBounds(selectedRegion.bounds, animated: true)
+        }
+
+        @objc func startOfflinePackDownload() {
+
+            // Create a region that includes the current viewport and any tiles needed to view it when zoomed further in.
+            // Because tile count grows exponentially with the maximum zoom level, you should be conservative with your `toZoomLevel` setting.
+            let region = MGLTilePyramidOfflineRegion(styleURL: mapView.styleURL, bounds: mapView.visibleCoordinateBounds, fromZoomLevel: mapView.zoomLevel, toZoomLevel: 14)
```
captainbarbosa, line 136: What if the user is already past z14, would this break?

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+            view.addSubview(mapView)
+            view.addSubview(tableView)
+            view.addSubview(downloadButton)
+            mapView.setCenter(CLLocationCoordinate2D(latitude: 22.27933, longitude: 114.16281),
+
+                              zoomLevel: 13, animated: false)
+
+            // Setup offline pack notification handlers.
+            NotificationCenter.default.addObserver(self, selector: #selector(offlinePackProgressDidChange), name: NSNotification.Name.MGLOfflinePackProgressChanged, object: nil)
+            NotificationCenter.default.addObserver(self, selector: #selector(offlinePackDidReceiveError), name: NSNotification.Name.MGLOfflinePackError, object: nil)
```
captainbarbosa, line 49: I don't think we need to watch these two notification types, since they're covered in the other example. We can just watch `MGLOfflinePackProgressChanged`.

captainbarbosa (changes requested):
I like how this ended up! Let me know if you have questions 👍

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
+import Mapbox
+
+@objc(ManageOfflineRegionsExample_Swift)
+
+class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, UITableViewDelegate, UITableViewDataSource  {
```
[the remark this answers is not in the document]
ZiZasaurus: added the delete swipe function! :)

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+            NotificationCenter.default.addObserver(self, selector: #selector(offlinePackDidReceiveError), name: NSNotification.Name.MGLOfflinePackError, object: nil)
+            NotificationCenter.default.addObserver(self, selector: #selector(offlinePackDidReceiveMaximumAllowedMapboxTiles), name: NSNotification.Name.MGLOfflinePackMaximumMapboxTilesReached, object: nil)
+            installConstraints()
+        }
+
+        func installConstraints() {
+
+            if #available(iOS 11.0, *) {
+                NSLayoutConstraint.activate([
+                    downloadButton.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
```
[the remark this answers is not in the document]
ZiZasaurus: no compass in this example but i'm happy to move over to the left hand side. Whatever you think looks best!

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -1,220 +1,234 @@
...
+            mapView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            mapView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
+            mapView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
+            mapView.heightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.heightAnchor, multiplier: 0.75),
+            tableView.topAnchor.constraint(equalTo: mapView.bottomAnchor),
+            tableView.leadingAnchor.constraint(equalTo: mapView.leadingAnchor),
+            tableView.trailingAnchor.constraint(equalTo: mapView.trailingAnchor),
+            tableView.heightAnchor.constraint(equalTo: mapView.heightAnchor, multiplier: 0.4),
+            downloadButton.topAnchor.constraint(equalTo: mapView.topAnchor, constant: 10),
+            downloadButton.leadingAnchor.constraint(equalTo: mapView.leadingAnchor),
```
captainbarbosa, line 75: Can we add some extra space here? Looks a little too close to the edge:

image

I'd also consider making the button a contrasting color with some bold text so it stands out from the map view's color scheme. Maybe try one of the [system colors](https://developer.apple.com/documentation/uikit/uicolor/standard_colors)?

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -1,220 +1,234 @@
...
+                let progress = completed / expected
+            cell.textLabel?.text = "Region \(indexPath.row + 1): size: \(packs[indexPath.row].progress.countOfBytesCompleted)"
+            cell.detailTextLabel?.text = "Percent completion: \(String(format: "%.2f", progress * 100))%"
+            cell.textLabel?.textAlignment = .left
+
+        } else {
+            print("so...")
+            cell.textLabel?.text = "No Offline Packs Saved"
+            cell.textLabel?.textColor = .darkGray
+            cell.textLabel?.textAlignment = .center
```
captainbarbosa, line 129: This would look nicer in a header. It would also be more efficient, since `cellForRowAt:indexPath` can potentially be called a lot of you have a lot of cells. If you already have at least one cell, then you know you've got packs, and so the rest of this if/else check is extra work.

`UITableViewDataSource` supports using titles to mark section headers. 

You can try something like:

```swift
func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
   /**
     If there are downloaded packs, return "Offline packs",
     otherwise return "No pack downloaded".
     */
}
```

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -1,220 +1,234 @@
...
+            return packs.count
+        } else {
+            return 1
+        }
+    }
 
-            return cell
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        var cell = tableView.dequeueReusableCell(withIdentifier: "cell", for: indexPath)
+        cell = UITableViewCell(style: .subtitle, reuseIdentifier: "cell")
```
captainbarbosa, line 116: `dequeueReusableCell(identifier:)` always returns a cell, so re-assigning it to a cell again is unnecessary.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -1,220 +1,234 @@
...
+
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        guard let packs = MGLOfflineStorage.shared.packs else { return }
+        if let selectedRegion = packs[indexPath.row].region as? MGLTilePyramidOfflineRegion {
             mapView.setVisibleCoordinateBounds(selectedRegion.bounds, animated: true)
         }
+    }
 
-        @objc func startOfflinePackDownload() {
+    @available(iOS 11.0, *)
```
captainbarbosa, line 144: I didn't realize this was an iOS 11+ things. Let's leave this part out since we don't want to have to deal with many `@available` checks in examples, as its not integral to the example being effective.

captainbarbosa (commented):
Let's review this in our next meetup

ZiZasaurus:
Worked with @captainbarbosa to add extension for progressPercentage calculation -- i added an if statement in the extension because it wouldn't run properly otherwise.

Also addressed the following issues: 
- why cells were duplicating
- cleaned up formatting and removed unnecessary print statements (left the ones from the original example)
- fixed cell subtitle issue

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -111,23 +94,28 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
...
+            return label.text
+        }
+
+        return nil
+    }
+
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        var cell = tableView.dequeueReusableCell(withIdentifier: "cell", for: indexPath)
+
         cell = UITableViewCell(style: .subtitle, reuseIdentifier: "cell")
```
captainbarbosa, line 111: Ohhh see what you mean now with the cell style thing. Here's a simply approach I think could work:

```swift
let cell = tableView.dequeueReusableCell(withIdentifier: "cell") ?? UITableViewCell(style: .subtitle, 
                                                                    reuseIdentifier: "cell")
```

This checks `dequeueReusableCell` to see if there is already a cell prepared with that identifier, and if there isn't, it creates a new cell with the `subtitle` style.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -111,23 +94,28 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
         }
     }
 
+    func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
```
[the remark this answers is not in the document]
ZiZasaurus: Yea I realized this and forgot to change it *facepalm*

Examples/Swift/ManageOfflineRegionsExample.swift
ZiZasaurus: changing the code to this still doesn't allow the subtitle to show :(

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -221,14 +178,30 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
...
-        }
+}
+
+fileprivate extension MGLOfflinePackProgress {
 
+    var progressPercentage: Float {
+        if countOfResourcesExpected != 0 {
+            return Float((countOfResourcesCompleted / countOfResourcesExpected) * 100)
+        } else {
+            return 1.0
```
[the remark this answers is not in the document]
ZiZasaurus: changed it to 0

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -221,14 +178,30 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
 
                 // Otherwise, print download/verification progress.
 
-                print("Offline pack “\(userInfo["name"] ?? "unknown")” has \(completedResources) of \(expectedResources) resources — \(String(format: "%.2f", progressPercentage! * 100))%.")
+                print("Offline pack “\(userInfo["name"] ?? "unknown")” has \(completedResources) of \(expectedResources) resources — \(String(format: "%.2f", pack.progress.progressPercentage))%.")
```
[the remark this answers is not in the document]
ZiZasaurus: nope not needed! removed!

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -202,14 +164,9 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
...
-            let progress = pack.progress
-            // or notification.userInfo![MGLOfflinePackProgressUserInfoKey]!.MGLOfflinePackProgressValue
-            let completedResources = progress.countOfResourcesCompleted
-            let expectedResources = progress.countOfResourcesExpected
 
-            // Calculate current progress percentage.
-
-            progressPercentage = Float(completedResources) / Float(expectedResources)
+            let completedResources = pack.progress.countOfResourcesCompleted
+            let expectedResources = pack.progress.countOfResourcesExpected
```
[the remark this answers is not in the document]
ZiZasaurus: removed!

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -111,23 +94,28 @@ class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate, U
...
         cell = UITableViewCell(style: .subtitle, reuseIdentifier: "cell")
+
         if let packs = MGLOfflineStorage.shared.packs {
-            let completed = Float(packs[indexPath.row].progress.countOfResourcesCompleted)
-                let expected = Float(packs[indexPath.row].progress.countOfResourcesExpected)
-                let progress = completed / expected
+
             cell.textLabel?.text = "Region \(indexPath.row + 1): size: \(packs[indexPath.row].progress.countOfBytesCompleted)"
-            cell.detailTextLabel?.text = "Percent completion: \(String(format: "%.2f", progress * 100))%"
+            cell.detailTextLabel?.text = "Percent completion: \(String(format: "%.2f", packs[indexPath.row].progress.progressPercentage))%"
```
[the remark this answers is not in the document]
ZiZasaurus: fixed!

captainbarbosa:
@ZiZasaurus please re-request review when ready (the little 🔁 symbol next to my name in the reviewers list)

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,190 @@
...
+        guard let packs = MGLOfflineStorage.shared.packs else { return }
+        if let selectedRegion = packs[indexPath.row].region as? MGLTilePyramidOfflineRegion {
+            mapView.setVisibleCoordinateBounds(selectedRegion.bounds, animated: true)
+        }
+    }
+
+    @objc func startOfflinePackDownload() {
+
+        // Create a region that includes the current viewport and any tiles needed to view it when zoomed further in.
+        // Because tile count grows exponentially with the maximum zoom level, you should be conservative with your `toZoomLevel` setting.
```
captainbarbosa, line 136:
```suggestion
        /**
         Create a region that includes the current map camera, to be captured
         in an offline map. Note: Because tile count grows exponentially as zoom level
         increases, you should be conservative with your `toZoomLevel` setting.
         */
```

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,190 @@
...
+        let region = MGLTilePyramidOfflineRegion(styleURL: mapView.styleURL, bounds: mapView.visibleCoordinateBounds, fromZoomLevel: mapView.zoomLevel, toZoomLevel: mapView.zoomLevel + 2)
+
+        // Store some data for identification purposes alongside the downloaded resources.
```
captainbarbosa, line 139:
```suggestion
        // Store some data for identification purposes alongside the offline pack.
```

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,190 @@
...
+    override func viewDidLoad() {
+
+        super.viewDidLoad()
+
+        view.addSubview(mapView)
+        view.addSubview(tableView)
+        mapView.addSubview(downloadButton)
+        mapView.setCenter(CLLocationCoordinate2D(latitude: 22.27933, longitude: 114.16281),
+
+                          zoomLevel: 13, animated: false)
```
captainbarbosa, line 44:
```suggestion
        let centerCoordinate = CLLocationCoordinate2D(latitude: 22.27933, longitude: 114.16281)
        mapView.setCenter(centerCoordinate, zoomLevel: 13, animated: false)
```

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,190 @@
...
+
+            if pack.progress.countOfResourcesCompleted == pack.progress.countOfResourcesExpected {
+
+                let byteCount = ByteCountFormatter.string(fromByteCount: Int64(pack.progress.countOfBytesCompleted), countStyle: ByteCountFormatter.CountStyle.memory)
+
+                print("Offline pack “\(userInfo["name"] ?? "unknown")” completed: \(byteCount), \(pack.progress.countOfResourcesCompleted) resources")
+
+            }
+        }
+        self.tableView.reloadData()
```
captainbarbosa, line 177:
```suggestion
        // Reload the table to update the progress percentage for each offline pack.
        self.tableView.reloadData()
```

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,190 @@
...
+
+    }
+
+}
+
```
captainbarbosa, line 182: Let's move all the methods that deal with setting up the `UITableView` here, into an extension that conforms to the protocols it needs:

```swift
extension ManageOfflineRegionsExample_Swift: UITableViewDelegate, UITableViewDataSource {
    // table view methods here
}
```

captainbarbosa (commented):
I've made some formatting suggestions which you can apply via [GitHub's suggested changes feature](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request#applying-a-suggested-change), and have a couple of minor code comments. Please also check line spacing as well - I was seeing a couple of SwiftLint warnings. This is close!

ZiZasaurus:
Git made me do a merge so let me know if anything looks off

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,197 @@
...
+            mapView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            mapView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            mapView.heightAnchor.constraint(equalTo: view.heightAnchor, multiplier: 0.75),
+            tableView.topAnchor.constraint(equalTo: mapView.bottomAnchor),
+            tableView.leadingAnchor.constraint(equalTo: mapView.leadingAnchor),
+            tableView.trailingAnchor.constraint(equalTo: mapView.trailingAnchor),
+            tableView.heightAnchor.constraint(equalTo: mapView.heightAnchor, multiplier: 0.4),
+            downloadButton.topAnchor.constraint(equalTo: mapView.topAnchor, constant: 100),
+            downloadButton.leadingAnchor.constraint(equalTo: mapView.leadingAnchor, constant: 5),
+            downloadButton.trailingAnchor.constraint(equalTo: mapView.leadingAnchor, constant: 150)
```
captainbarbosa, line 69: I think there are some improvements that can be made here -

**Map view / table view**:

The only constraints here that should be relative to the `mapView` are the table view's top. The rest can just be relative to the `view` itself.

<details>
<summary>OR, you can take on a ✨ bonus challenge ✨... </summary>

<br>

The map view and table view take up almost 50/50 in height. Instead of manually setting the constrains for these views, I challenge you to experiment with placing both of these in a [`UIStackView`](https://developer.apple.com/documentation/uikit/uistackview). The `UIStackView` will need to have its `distribution` and `axis` properties set to the correct values for this to work.
</details>

**Download button**:

Instead of doing a trailing anchor for the download button, let's set a `width` constraint instead. I'd also add some `height` since the button looks a little squished:

image


To fit with the design and to stand out from the map more, can you reverse the button's text color and background color?

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,197 @@
...
+            cell.detailTextLabel?.text = "Percent completion: \(pack.progress.percentCompleted)%"
+
+        }
+
+        return cell
+
+    }
+
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        guard let packs = MGLOfflineStorage.shared.packs else { return }
```
captainbarbosa, line 192: Just adding an extra space below this line:

```suggestion
        guard let packs = MGLOfflineStorage.shared.packs else { return }
        
```

captainbarbosa (changes requested):
Added some formatting suggestions, I think the UI needs a little cleanup and then we should be good to go 🚀 

Also, I'd be fine with the Obj-C version coming as a follow up PR. This PR is already a little big.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+            }
+        }
+        // Reload the table to update the progress percentage for each offline pack.
+        self.tableView.reloadData()
+
+    }
+
+}
+
+fileprivate extension MGLOfflinePackProgress {
```
captainbarbosa, line 164: Lets move this below the table view methods.

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+
+        let cell = UITableViewCell.init(style: .subtitle, reuseIdentifier: "cell")
+
+        if let packs = MGLOfflineStorage.shared.packs {
+            let pack = packs[indexPath.row]
+
+            cell.textLabel?.text = "Region \(indexPath.row + 1): size: \(pack.progress.countOfBytesCompleted)"
```
captainbarbosa, line 211: I think we should convert `pack.progress.countOfBytesCompleted` to MB, such that 600000 bytes could be transformed to a string such as `0.6 MB`.

Check out [`ByteCountFormatter`](https://developer.apple.com/documentation/foundation/bytecountformatter) which helps you do this. You'll want to create the `ByteCountFormatter` _outside_ of this method (so in a global constant) so it only gets created once for the entire table instead of for a per-cell basis, and then call the appropriate method to format the value here.

If you don't end up implementing this, I'd at least add the units to the end of this string as the user doesn't know what unit value is being displayed.

captainbarbosa (approved):
✅ pending some minor comments. Nice work!

Examples/Swift/ManageOfflineRegionsExample.swift
captainbarbosa: (Meaning, move `fileprivate extension MGLOfflinePackProgress {...}` below all the table view methods)

Examples/Swift/ManageOfflineRegionsExample.swift
```diff
@@ -5,6 +5,8 @@ import Foundation
 
 class ManageOfflineRegionsExample_Swift: UIViewController, MGLMapViewDelegate {
 
+    let inBytes = ByteCountFormatter()
```
captainbarbosa, line 8: I accidentally led you astray with this suggestion, I'm sorry! 

So, it looks like the `ByteCountFormatter.string(fromByteCount:countStyle:)` is a _class_ method. This means we don't need to initialize a `ByteCountFormatter` to call this specific method:

For example,

```swift
ByteCountFormatter().string(fromByteCount:countStyle:) // won't work
ByteCountFormatter.string(fromByteCount:countStyle:) // works
```

Because of this, there actually isn't a need to hold a reference to a `ByteCountFormatter` here, because we don't need an instance of the `ByteCountFormatter` class to call the method we want to use. 

In this file you're using two different methods on `ByteCountFormatter`. One is the class function (or class method), and the other is an instance method:

```swift
// Within offlinePackProgressDidChange
// Class method - called on the type itself
let byteCount = ByteCountFormatter.string(fromByteCount: Int64(pack.progress.countOfBytesCompleted), countStyle: ByteCountFormatter.CountStyle.memory)

// Within cellForRowAt:indexPath
// Instance method - called on an instance of the type
cell.textLabel?.text = "Region \(indexPath.row + 1): size: \(inBytes.string(fromByteCount: Int64(pack.progress.countOfBytesCompleted)))"
```

Here's a suggestion - how about we add a second computed property on `MGLOfflinePackProgress`:

```swift
var formattedCountOfBytesCompleted: String {
    return ByteCountFormatter.string(fromByteCount: Int64(countOfBytesCompleted),
                                     countStyle: .memory)
}
```

And then the updates will look like this:

```swift
// Within offlinePackProgressDidChange
let byteCount = pack.progress.formattedCountOfBytesCompleted

// Within cellForRowAt:indexPath
cell.textLabel?.text = "Region \(indexPath.row + 1): size: \(pack.progress.formattedCountOfBytesCompleted)"
```

captainbarbosa (approved):
Please merge after reviewing the last suggestion 👍 🏆

PR #421, opened 2020-11-17
Updating existing offline example to fix #419

### LINKED ISSUE #419: Completed Resources test for Offline Pack fires more than once in `OfflinePackExample.swift`
In `OfflinePackExample.swift`, when downloading an offline pack there is a test to determine whether the _Completed Resources is equal to the Expected Resources_.

The test that determines _download complete_ fires more than once, and I'm looking for suggestions on how to improve handling this.

* Idea # 1:  Add an instance variable, `allDownloaded` or whatever,  in the class `OfflinePackExample_Swift` that tracks the first time that `completedResources == expectedResources`, and never enter that clause again.
* Fix on the Mapbox GL native side of the code
* Fixes need to be added to ObjC sample as well
* File this issue under _improve existing examples._
* Goal of this issue is to request an update to the Mapbox Examples code

https://github.com/mapbox/ios-sdk-examples/blob/5fff10393e0d48f0fdc2b98e1b0be199feb8e017/Examples/Swift/OfflinePackExample.swift#L101,L103

---

In practice, we want to use the state that _all expected resources have downloaded_, and update a user interface.  In our app, the test for `completedResources == expectedResources` is true more than twice.

This issues when occurs downloading from the following apps

* The app in this repo, _Examples_, using iOS SDK 6.2.1 and creating an offline pack from Mapbox dark style
* Our app, using iOS SDK 5.6.1 and our custom style

---

Log from _Examples_, where the `completed` state is logged twice.
```bash 
Offline pack “My Offline Pack” has 4 of 545 resources — 0.73%.
Offline pack “My Offline Pack” has 69 of 545 resources — 12.66%.
... <snip a bunch>
Offline pack “My Offline Pack” has 544 of 545 resources — 99.82%.
Offline pack “My Offline Pack” completed: 7 MB, 545 resources
Offline pack “My Offline Pack” completed: 7 MB, 545 resources
```

---

Our Apps' offline packs support downloading non-contiguous grids.  In our case, our UI reports that _4 out of 2_ grids have been downloaded.

image

ZiZasaurus:
Fixes https://github.com/mapbox/ios-sdk-examples/issues/419

### DIFF
--- a/Examples/Swift/OfflinePackExample.swift
+++ b/Examples/Swift/OfflinePackExample.swift
@@ -18,7 +18,9 @@ class OfflinePackExample_Swift: UIViewController, MGLMapViewDelegate {
         mapView.setCenter(CLLocationCoordinate2D(latitude: 22.27933, longitude: 114.16281),
                           zoomLevel: 13, animated: false)
 
-        // Setup offline pack notification handlers.
+    }
+
+    func setupOfflineNotificationHandlers() {
         NotificationCenter.default.addObserver(self, selector: #selector(offlinePackProgressDidChange), name: NSNotification.Name.MGLOfflinePackProgressChanged, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(offlinePackDidReceiveError), name: NSNotification.Name.MGLOfflinePackError, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(offlinePackDidReceiveMaximumAllowedMapboxTiles), name: NSNotification.Name.MGLOfflinePackMaximumMapboxTilesReached, object: nil)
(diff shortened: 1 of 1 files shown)


Examples/Swift/OfflinePackExample.swift
```diff
@@ -42,13 +44,10 @@ class OfflinePackExample_Swift: UIViewController, MGLMapViewDelegate {
         }
     }
 
-    deinit {
-        // Remove offline pack observers.
-        print("Removing offline pack notification observers")
-        NotificationCenter.default.removeObserver(self)
```
1ec5, line 48: It’s still important to clean up observations when deinitializing (or when the view controller is about to disappear, depending). The application may allow the user to navigate away from the offline pack–downloading view controller before the download completes. If that happens, the view controller shouldn’t leak memory due to a lingering observation.

Examples/Swift/OfflinePackExample.swift
```diff
@@ -101,6 +100,12 @@ class OfflinePackExample_Swift: UIViewController, MGLMapViewDelegate {
             if completedResources == expectedResources {
                 let byteCount = ByteCountFormatter.string(fromByteCount: Int64(pack.progress.countOfBytesCompleted), countStyle: ByteCountFormatter.CountStyle.memory)
                 print("Offline pack “\(userInfo["name"] ?? "unknown")” completed: \(byteCount), \(completedResources) resources")
+
+                // Remove notification handlers
+                NotificationCenter.default.removeObserver(self, name: NSNotification.Name.MGLOfflinePackProgressChanged, object: nil)
+                NotificationCenter.default.removeObserver(self, name: NSNotification.Name.MGLOfflinePackError, object: nil)
+                NotificationCenter.default.removeObserver(self, name: NSNotification.Name.MGLOfflinePackMaximumMapboxTilesReached, object: nil)
```
1ec5, line 107: Each of these observations would be worth removing during deinitialization too. Similar to what you did with the calls to `addObserver(_:selector:name:object:)`, you can factor these calls to `removeObserver(_:name:object:)` into a separate method for consistency.

ZiZasaurus: oddly enough, changing `completedResources == expectedResources` to `pack.state == .complete` resolves the duplicate notification in this example

PR #422, opened 2020-12-09
Add objective c example for managing offline regions

ZiZasaurus:


Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+- (void)startOfflinePackDownload: (UIButton *)sender {
+    // Setup offline pack notification handlers.
+    [self setupOfflinePackHandler];
+
+    /**
+     Create a region that includes the current map camera, to be captured
+     in an offline map. Note: Because tile count grows exponentially as zoom level
+     increases, you should be conservative with your `toZoomLevel` setting.
+     */
+    id <MGLOfflineRegion> region = [[MGLTilePyramidOfflineRegion alloc] initWithStyleURL:self.mapView.styleURL bounds:self.mapView.visibleCoordinateBounds fromZoomLevel:self.mapView.zoomLevel toZoomLevel: self.mapView.zoomLevel + 2];
```
captainbarbosa, line 142: The convention to define variables in ObjC goes like:

```
<object type> <optional pointer><variable name> = <assignment>;
```

(The pointer `*` is only sometimes required depending on the type - Xcode will give you a helpful warning if its missing or needed though.)

In this case, you're specifying that `region` is going to be an `id` of conforming to the MGLOfflineRegion protocol. A cleaner way to do this would be to just use the type in the definition already:


```objc
MGLTilePyramidOfflineRegion *region = [[MGLTilePyramidOfflineRegion alloc] initWithStyleURL:self.mapView.styleURL
                                                                                     bounds:self.mapView.visibleCoordinateBounds
                                                                              fromZoomLevel:self.mapView.zoomLevel
                                                                                toZoomLevel: self.mapView.zoomLevel + 2];
```

In addition, in Obj-C the formatting convention is to align parameters by the colon `:` along new lines, as shown above.

Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+heightForHeaderInSection:(NSInteger)section {
+    return 50.0;
+}
+
+- (nonnull UITableViewCell *)tableView:(nonnull UITableView *)tableView
+                 cellForRowAtIndexPath:(nonnull NSIndexPath *)indexPath {
+
+    UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"cell" forIndexPath:indexPath];
+
+    [cell initWithStyle:UITableViewCellStyleSubtitle reuseIdentifier:@"cell"];
```
captainbarbosa, line 251: The cell is being initialized twice here. Instead, you can do this:

```suggestion
    UITableViewCell *cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleSubtitle
                                                 reuseIdentifier:@"cell"];
```

(Note - I also aligned the function call along its parameter colons)

Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+    
+    if (MGLOfflineStorage.sharedOfflineStorage.packs.count != 0 )  {
+        MGLOfflinePack * pack = MGLOfflineStorage.sharedOfflineStorage.packs[indexPath.row];
+        long number = indexPath.row + 1;
+        long countOfBytesCompleted = MGLOfflineStorage.sharedOfflineStorage.packs[indexPath.row].progress.countOfBytesCompleted;
+        NSByteCountFormatter *formattedCountOfBytedCompleted = [NSByteCountFormatter stringFromByteCount:countOfBytesCompleted countStyle:NSByteCountFormatterCountStyleMemory];
```
captainbarbosa, line 257: Here I'm seeing a warning:

```
Incompatible pointer types initializing 'NSByteCountFormatter *' 
with an expression of type 'NSString * _Nonnull'
```

This is because the method `[NSByteCountFormatter stringFromByteCount:countStyle:]` returns `NSString`, and not `NSByteCountFormatter` which is defined before the variable name.

Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+- (nonnull UITableViewCell *)tableView:(nonnull UITableView *)tableView
+                 cellForRowAtIndexPath:(nonnull NSIndexPath *)indexPath {
+
+    UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"cell" forIndexPath:indexPath];
+
+    [cell initWithStyle:UITableViewCellStyleSubtitle reuseIdentifier:@"cell"];
+    
+    if (MGLOfflineStorage.sharedOfflineStorage.packs.count != 0 )  {
+        MGLOfflinePack * pack = MGLOfflineStorage.sharedOfflineStorage.packs[indexPath.row];
+        long number = indexPath.row + 1;
```
captainbarbosa, line 255: Suggesting a switch from `long` to `NSInteger`.

```suggestion
        NSInteger number = indexPath.row + 1;
```

Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+        long countOfBytesCompleted = MGLOfflineStorage.sharedOfflineStorage.packs[indexPath.row].progress.countOfBytesCompleted;
```
captainbarbosa, line 256:
```suggestion
        NSInteger countOfBytesCompleted = MGLOfflineStorage.sharedOfflineStorage.packs[indexPath.row].progress.countOfBytesCompleted;
```

Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+        NSByteCountFormatter *formattedCountOfBytedCompleted = [NSByteCountFormatter stringFromByteCount:countOfBytesCompleted countStyle:NSByteCountFormatterCountStyleMemory];
+        float countOfResourcesCompleted = pack.progress.countOfResourcesCompleted;
```
captainbarbosa, line 258: Lets match the return type that this property defines - `countOfResourcesCompleted` is of type `UInt64`
```suggestion
        UInt64 countOfResourcesCompleted = pack.progress.countOfResourcesCompleted;
```

Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+        float countOfResourcesExpected = pack.progress.countOfResourcesExpected;
```
captainbarbosa, line 259:
```suggestion
        UInt64 countOfResourcesExpected = pack.progress.countOfResourcesExpected;
```

Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+    }
+
+    return cell;
+}
+
+-(void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath {
+    MGLTilePyramidOfflineRegion *selectedRegion = MGLOfflineStorage.sharedOfflineStorage.packs[indexPath.row].region;
+
+    if ([selectedRegion isKindOfClass:[MGLTilePyramidOfflineRegion class]]) {
+        [_mapView setVisibleCoordinateBounds:selectedRegion.bounds animated:YES];
```
captainbarbosa, line 276: Please use the `self.mapView` style of setting all properties instead of `_mapView`, which is the syntax we use across all ObjC examples. (You can read more about [here](https://stackoverflow.com/questions/14112715/self-variablename-vs-variablename-vs-sysnthesize-variablename) and [here](https://teamtreehouse.com/community/objc-underscore-self-notation-difference)).

This should also allow you to remove all your `@synthesize` calls.

captainbarbosa (changes requested):

Examples/ObjectiveC/ManageOfflineRegionsExample.m
```diff
@@ -3,12 +3,281 @@
...
+@property (nonatomic) UIButton *downloadButton;
+@property (nonatomic) UITableView *tableView;
 @property (nonatomic) UIProgressView *progressView;
 @end
 
 @implementation ManageOfflineRegionsExample
 
+@synthesize mapView = _mapView;
+
+- (id)mapView
```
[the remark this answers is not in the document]
ZiZasaurus: ah perfect, thanks for this!

Examples/ObjectiveC/ManageOfflineRegionsExample.m
ZiZasaurus: Thanks! When I try to align the region by colon, it looks WILD 😩

Examples/ObjectiveC/ManageOfflineRegionsExample.m
ZiZasaurus: 🤦🏾‍♀️, thank you!

captainbarbosa (approved):
✅ hopefully this is the last ObjC example we have to build for a while 😄

